PR #29, opened 2025-10-19
feat: test all registered FWSS providers, not just approved ones

pali101:
Related to FilOzone/filecoin-services#302

This PR refactors the dealbot to align with the new permissionless, self-service model of the FilecoinWarmStorageService contract.

Instead of only testing the small `approved` list, the bot now loads, creates deals for, and runs retrieval tests against all providers registered in the `ServiceProviderRegistry`. This is a crucial change to enable performance scoring for all providers who wish to join the service.

## Key changes

- `WalletSdkService`:
   - Now fetches all providers into `registeredProviders` on init.
   - Tracks `approved` status via a new `approvedProviderIds` `Set` for filtering.
   - Logs now show the full breakdown (e.g., "Loaded 22 registered providers (7 approved, 15 not approved)").
   
- `DealService` and `SchedulerService`:
   - Updated to iterate over the full `registeredProviders` list for deal creation.
   - Logs now correctly refer to testing all registered providers.

- `RetrievalService`:
   - Switched from `getApprovedProviderInfo()` to `getProviderInfo()` to allow retrievals from any registered provider.

## Breaking change

`WalletSdkService.getProviderCount()` now returns the count of all registered providers, not just approved ones.

For the old behavior, use the new `getApprovedProviderCount()` method.

### DIFF
--- a/src/wallet-sdk/wallet-sdk.service.ts
+++ b/src/wallet-sdk/wallet-sdk.service.ts
@@ -87,7 +87,7 @@ export class WalletSdkService implements OnModuleInit {
       this.registeredProviders = validProviders.filter(provider => {
         const isActive = provider.active;
         const supportsPDP = !!provider.products?.PDP;
-        const isDevTagged = provider.products?.PDP?.capabilities?.serviceStatus === 'dev';
+        const isDevTagged = provider.products?.PDP?.capabilities?.service_status === 'dev';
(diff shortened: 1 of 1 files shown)


Kubuxu:
We discussed in the daily sync. We should not test providers tagged with `dev`.

PR #52, opened 2025-11-06
Feat: add ipfs retrieval checks

### LINKED ISSUE #51: Add IPFS tooling-based retrieval checks alongside direct-to-SP retrieval
Dealbot currently only tests direct-to-SP piece retrieval, as well as CDN retrievals. But we do not catch IPNI announcement failures.

Would it be feasible to implement retrieval checks using IPFS tooling (e.g., via IPNI lookups + standard IPFS retrieval). This would help catch SPs that can serve pieces directly but aren't properly announcing to IPNI, which impacts broader network retrievability.

silent-cipher:
Failed Ipfs retrieval details can be seen in failed retrievals section.
Preview - 
Screenshot 2025-11-06 at 1 31 48 PM

closes #51

### DIFF
--- a/src/deal-addons/strategies/ipni.strategy.ts
+++ b/src/deal-addons/strategies/ipni.strategy.ts
@@ -229,6 +229,8 @@ export class IpniAddonStrategy implements IDealAddon<IpniMetadata> {
           advertised: false,
           retrieved: false,
           retrievedAt: null,
+          indexedAt: null,
+          advertisedAt: null,
         },
         checks: 0,
         durationMs: statusTimeoutMs,
@@ -263,6 +265,8 @@ export class IpniAddonStrategy implements IDealAddon<IpniMetadata> {
       advertised: false,
       retrieved: false,
       retrievedAt: null,
+      indexedAt: null,
+      advertisedAt: null,
     };
     let checkCount = 0;
 
@@ -270,27 +274,40 @@ export class IpniAddonStrategy implements IDealAddon<IpniMetadata> {
       checkCount++;
 
       try {
-        const status = await pdpServer.getPieceStatus(pieceCid);
+        const sdkStatus = await pdpServer.getPieceStatus(pieceCid);
 
-        // Log state transitions
-        if (status.indexed && !lastStatus.indexed) {
+        const currentStatus: PieceStatus = {
+          status: sdkStatus.status,
+          indexed: sdkStatus.indexed,
+          advertised: sdkStatus.advertised,
+          retrieved: sdkStatus.retrieved,
+          retrievedAt: sdkStatus.retrievedAt,
+          indexedAt: lastStatus.indexedAt,
+          advertisedAt: lastStatus.advertisedAt,
+        };
+
+        if (currentStatus.indexed && !lastStatus.indexed) {
+          currentStatus.indexedAt = new Date().toISOString();
           this.logger.log(`Piece indexed: ${pieceCid.slice(0, 12)}...`);
         }
-        if (status.advertised && !lastStatus.advertised) {
+
+        if (currentStatus.advertised && !lastStatus.advertised) {
+          currentStatus.advertisedAt = new Date().toISOString();
           this.logger.log(`Piece advertised: ${pieceCid.slice(0, 12)}...`);
         }
-        if (status.retrievedAt && !lastStatus.retrievedAt) {
+
+        if (currentStatus.retrievedAt && !lastStatus.retrievedAt) {
           const durationSec = ((Date.now() - startTime) / 1000).toFixed(1);
           this.logger.log(`Piece retrieved: ${pieceCid.slice(0, 12)}... (${durationSec}s)`);
           return {
             success: true,
-            finalStatus: status,
+            finalStatus: currentStatus,
             checks: checkCount,
             durationMs: Date.now() - startTime,
           };
         }
 
-        lastStatus = status;
+        lastStatus = currentStatus;
       } catch (error) {
         if (checkCount % 20 === 0) {
           this.logger.debug(`Status check error: ${error.message}`);
@@ -349,6 +366,7 @@ export class IpniAddonStrategy implements IDealAddon<IpniMetadata> {
     }
 
     const durationMs = Date.now() - startTime;
+    const verifiedAt = new Date().toISOString();
     const successCount = (rootCIDResult.verified ? 1 : 0) + blockCIDsVerified;
     const totalCIDs = blockCIDs.length + 1; // +1 for rootCID
 
@@ -366,6 +384,7 @@ export class IpniAddonStrategy implements IDealAddon<IpniMetadata> {
       rootCIDVerified: rootCIDResult.verified,
       durationMs,
       failedCIDs,
+      verifiedAt,
     };
   }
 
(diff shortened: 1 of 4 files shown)


src/database/entities/sp-performance-all-time.entity.ts
```diff
@@ -31,14 +31,14 @@ import { ServiceType } from "../types.js";
         -- Deal throughput (all time)
         ROUND(AVG(d.ingest_throughput_bps) FILTER (WHERE d.ingest_throughput_bps IS NOT NULL))::bigint as avg_ingest_throughput_bps,
         
-        -- Retrieval metrics (all time)
+        -- Retrieval metrics (all time - DIRECT_SP only)
         COUNT(DISTINCT r.id) FILTER (WHERE r.service_type = '${ServiceType.DIRECT_SP}') as total_retrievals,
         COUNT(DISTINCT r.id) FILTER (WHERE r.status = 'success' AND r.service_type = '${ServiceType.DIRECT_SP}') as successful_retrievals,
         COUNT(DISTINCT r.id) FILTER (WHERE r.status = 'failed' AND r.service_type = '${ServiceType.DIRECT_SP}') as failed_retrievals,
         
         -- Retrieval success rate (all time)
```
BigLep:
```suggestion
        -- Direct SP /piece Retrieval success rate (all time)
```

src/database/entities/sp-performance-all-time.entity.ts
```diff
@@ -56,19 +56,67 @@ import { ServiceType } from "../types.js";
         -- Retrieval throughput (all time)
         ROUND(AVG(r.throughput_bps) FILTER (WHERE r.throughput_bps IS NOT NULL AND r.service_type = '${ServiceType.DIRECT_SP}'))::bigint as avg_throughput_bps,
         
+        -- IPFS retrieval metrics (all time - IPFS_PIN only)
+        COUNT(DISTINCT r_ipfs.id) FILTER (WHERE r_ipfs.service_type = '${ServiceType.IPFS_PIN}') as total_ipfs_retrievals,
+        COUNT(DISTINCT r_ipfs.id) FILTER (WHERE r_ipfs.status = 'success' AND r_ipfs.service_type = '${ServiceType.IPFS_PIN}') as successful_ipfs_retrievals,
+        COUNT(DISTINCT r_ipfs.id) FILTER (WHERE r_ipfs.status = 'failed' AND r_ipfs.service_type = '${ServiceType.IPFS_PIN}') as failed_ipfs_retrievals,
+        
+        -- IPFS retrieval success rate
```
BigLep: nit (not that you need to change) , but my understanding is that "success rates" are between 0 and 1 while "success percentages" are between 0 and 100.  In this case we are calling a "success percentage" a "success rate".

BigLep (commented):
Thanks for putting this together.  Lots to pull together and very important.  I'll respond separately on the screenshot itself.  Happy to talk verbally on any of this if it's useful.

## Concern 1: terminal state should be "verified" not "retrieved".
We're putting a lot of reliance on the PDPServer status of "retrieved".  

Per the [docs](https://github.com/FilOzone/synapse-sdk/blob/681df95bd376726fe403af2dd881db569af3227c/packages/synapse-sdk/src/pdp/server.ts#L106-L109):
```
   * Whether the piece has been retrieved
   * This does not necessarily mean it was retrieved by a particular indexer,
   * only that the PDP server witnessed a retrieval event. Care should be
   * taken when interpreting this field.
```

I think it's fine to include measurements (counts/latencies) around this "retrieved" stat, but I don't think we should treat it as a our terminal state for IPNI checking.  IPNI checking is done when we've verified CIDs are showing up in filecoinpin.contact.  

## Concern 2: state names
I would like to make the state names more self-documenting:
* indexed → sp_indexed ?: the SP claimed they indexed the piece
* advertised → sp_announced_advertisement : the SP announced a new advertisement chain to the indexer.  
   * That said, I realize PDPServer calls this "advertised", so I can see if we we want to be consistent there.  In that case, "sp_advertised".
* retrieved → sp_received_retrieve_request OR indexer_retrieved_from_sp
   * I think I prefer the first from an accuracy regard. I don't think we actually know if the Indexer retrieved the CIDs from the SP.
 * (add this) verified: meaning that we confirmed CIDs are actually being returned by the IPNI indexer (filecoinpin.contact).
    * A key decision point is when we can consider a "deal" verified.  Ideas:
       1. When the ipfsRootCid shows up properly in the IPNI response
       2. When more than a certain percentage of CIDs show up properly
       3. When all the CIDs have been verified
       * I think we definitely need to check the ipfsRootCID at the minimum.  Beyond that is bonus in my opinion. 

(do whatever you want with the casing - I'm just trying to convey an idea)

## Concern 3: transitioning between "sp_received_retrieve_request" and verifying with IPNI
Right now, as soon as we get to the "sp_received_retrieve_request" state we immediately start trying to verify IPNI results (see `monitorAndVerifyIPNI`).  That makes sense, but we should intentionally allow there to be some time between these two states.  Right now in `verifyIPNIAdvertisement` we just start iterating over each CID and marking them failed if we don't get a valid response.  In this asynchronous system, there may be some time between when CIDs are retrieved from the SP by the IPNI indexer and before they start showing up in IPNI responses.

BigLep:
Concerning
Screenshot 2025-11-06 at 1 31 48 PM

## Item 1: terminology
"Retrieval Success" is ambiguous to me.  Whenever we talk about "retrieval", I think we need to qualify it.

In my mind, we either have:
- "SP /piece Retrieval"
- "IPFS Mainnet Retrieval"

## Item 2: different success rates
I'm confused why we have one success rate above that is a bar and the 7-day success rate below just as a number.

I assume the top "Success Rate" is for all time.

Here is an idea for combining this to just using numbers rather than some numbers and some bars.

  | All Time Attempts | All Time Success Rate | 7 Day Attempts | 7 Day Success Rate
-- | -- | -- | -- | --
Uploads | 50 | 100% | 10 | 100%
SP /piece Retrieval | 50 | 100% | 10 | 100%
IPNI Indexing | 50 | 80% | 10 | 90%
IPFS Mainnet Retrieval | 50 | 70% | 10 | 90%

You don't have to go with this, but lets at least label clearly and be intentional on our organizing.  More on that below.

## Mixing Success Rates and Latency Metrics

The screenshot is odd to me right with the flow of:
- Success Rate (I assume all-time)
- IPNI & IPFS Metrics: here we are covering success rates and but also very latencies
- 7 Day Retrieval Success rate

I think either group by metric type (e.g., success rates then latency metrics) or group by system (Direct SP interactions, IPNI, IPFS Mainnet).  Right now you're mixing them.

My recommendation would be to move the IPNI and IPFS latency metrics down to the Latency section where you cover other latencies.

## Define the metric

The only state that is self-describing is "success rate".  I know how to combining multiple data points for a success rate.  But how are you combining:
- latencies
- TTFB's
- Time to Index
- Time to Advertise
etc.

Are you averaging, summing, counting, p50, etc.  Lets label to make it clear.

---
Again, feel free to pick through the feedback and I'm happy to talk verbally if that's helpful.  Thanks for your work here.

BigLep:
@silent-cipher : we've also learned that it's important that the https://sp.domain/ipfs endpoint support http2.  When doing the URL check for `/ipfs` can you make sure to use an client that doesn't have HTTP1 support?  Boxo (a Go IPFS client requires HTTP2 per https://github.com/ipfs/boxo/blob/fc4c0e6a90715d788462e715e8543a41abb17491/bitswap/network/httpnet/httpnet.go#L613-L621 )

silent-cipher:
Thanks for clarifying!
I had the same question: since the Go IPFS (Boxo) client requires HTTP/2, doesn’t that mean an SP that doesn’t support HTTP/2 wouldn’t work for `/ipfs` retrievals anyway?
If that’s the case, HTTP/2 support seems like a strict requirement rather than something we’d need to check dynamically for each request.

I’ve verified that the datacenter proxies I’m using do support HTTP/2, so that shouldn’t be an issue.
That said, I agree it might still make sense to do a one-time or periodic check for HTTP/2 support per SP instead of checking on every retrieval.

PR #67, opened 2025-12-01
feat: auto-fund and low balance alerts

pali101:
## Description

This PR introduces a mechanism to monitor the Dealbot's Filecoin-Pay (USDFC) balance, providing automatic top-ups and low-balance alerts to prevent deal failures.

## Key Changes

- Configuration: Added `WALLET_` and `ALERT_` env variables for thresholds, funding amounts, and webhooks.

- Wallet Logic: Implemented `checkAndHandleBalance` in `WalletSdkService`:
  - Checks USDFC balance against threshold.
  - Safety: Verifies `native FIL` gas and `USDFC` allowance before attempting deposits.
  - Concurrency: Uses an in-memory lock and mandatory cooldown (`WALLET_COOLDOWN_MINUTES`) to prevent race conditions and alert spam.

- Alerting: Added `AlertService` with retry logic to send generic webhook notifications (Slack-compatible).

- Scheduling: Added a dedicated cron job in `SchedulerService` to run the monitor periodically.

### New Environment Variables

```shell
WALLET_BALANCE_CHECK_INTERVAL_SECONDS=300
WALLET_BALANCE_THRESHOLD_USDFC=1000000000000000000  # 1 USDFC
WALLET_AUTO_FUND_AMOUNT_USDFC=500000000000000000    # 0.5 USDFC
WALLET_AUTO_FUND_ENABLED=true
WALLET_ALERT_ONLY_MODE=false
WALLET_COOLDOWN_MINUTES=30
ALERT_WEBHOOK_URL=https://hooks.slack.com/...
```

### DIFF
--- a/src/wallet-sdk/alert.service.ts
+++ b/src/wallet-sdk/alert.service.ts
@@ -126,10 +126,10 @@ export class AlertService {
         const statusCode = this.extractStatusCode(err);
 
         if (isLastAttempt) {
-          this.logger.error(
-            `Alert webhook failed after ${maxRetries} attempts (${context}); giving up`,
-            { error: String(err), statusCode },
-          );
+          this.logger.error(`Alert webhook failed after ${maxRetries} attempts (${context}); giving up`, {
+            error: String(err),
+            statusCode,
+          });
         } else {
           const delayMs = initialDelayMs * 2 ** attempt; // Exponential backoff: 1s, 2s, 4s
           this.logger.warn(
--- a/src/app.controller.ts
+++ b/src/app.controller.ts
@@ -37,16 +37,16 @@ export class AppController {
 
   /**
    * Debug webhook endpoint for testing alert service locally
-   * 
+   *
    * This endpoint is useful for development and testing without requiring
    * external webhook services (e.g., webhook.site, Discord, Slack).
-   * 
+   *
    * Usage:
    * Set in .env: ALERT_WEBHOOK_URL=http://localhost:8080/api/debug/webhook
-   * 
+   *
    * The endpoint logs all received alerts and returns a success response,
    * allowing you to verify alert payloads without rate limits.
-   * 
+   *
    * Note: Only intended for development/testing environments.
    */
   @Post("debug/webhook")


src/config/app.config.ts
```diff
@@ -132,6 +156,14 @@ export function loadConfig(): IConfig {
       retrievalStartOffsetSeconds: Number.parseInt(process.env.RETRIEVAL_START_OFFSET_SECONDS || "600", 10),
       metricsStartOffsetSeconds: Number.parseInt(process.env.METRICS_START_OFFSET_SECONDS || "900", 10),
     },
+    walletMonitor: {
+      balanceCheckIntervalSeconds: Number.parseInt(process.env.WALLET_BALANCE_CHECK_INTERVAL_SECONDS || "300", 10),
+      balanceThresholdUsdfc: process.env.WALLET_BALANCE_THRESHOLD_USDFC || "0",
+      autoFundAmountUsdfc: process.env.WALLET_AUTO_FUND_AMOUNT_USDFC || "0",
```
SgtPooki: why do we need to set the defaults in joi schema and here? can that all be centralized somewhere?

src/wallet-sdk/alert.service.ts
```diff
@@ -0,0 +1,51 @@
...
+    const payload = { type: "low_balance", details };
+    await this.postWithRetry(payload, "low_balance");
+  }
+
+  async sendFundResultAlert(details: Json): Promise<void> {
+    if (!this.webhookUrl) {
+      this.logger.warn("Alert webhook URL not configured; skipping fund result alert", details);
+      return;
+    }
+    const payload = { type: "auto_fund_result", details };
```
beck-8: Can you reconfirm that this structure is compatible with Slack?

src/config/app.config.ts
```diff
@@ -76,6 +85,19 @@ export interface ISchedulingConfig {
   metricsStartOffsetSeconds: number;
 }
 
+export interface IWalletMonitorConfig {
+  balanceCheckIntervalSeconds: number;
+  balanceThresholdUsdfc: string; // big integer string in smallest unit
+  autoFundAmountUsdfc: string; // big integer string in smallest unit
```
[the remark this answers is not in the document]
pali101: Valid point. Counting 18 zeros is error-prone.

I will update the config loader to accept human-readable units (e.g., 10 or 0.5) in the .env file and handle the conversion to BigInt (18 decimals) internally.

src/config/app.config.ts
pali101: I will clean this up to ensure the defaults are defined in a single source of truth.

pali101:
> Concerning the alarming, I'm wondering if we need to think more holistically here. For example, we probably want Grafana around this system anyways. If we have that, we can definer alarms there, and then configure the systems we want to alert (e.g., Slack).

Agreed, Grafana would be much better for long-term, holistic monitoring.

I think we should treat this PR as an intermediate solution to get basic alerts running now. I’m happy to take on the Grafana setup in a subsequent PR - I haven't used it before, but I'm eager to learn and implement it.

However, Grafana monitors, it doesn't execute. Since the application needs to monitor the balance internally to trigger the auto-funding transactions (logic that must exist within the app to sign the tx), these application-level alerts are a low-cost addition to that existing loop.

beck-8:
@pali101  Looks fine
image

.env.example
```diff
@@ -42,4 +42,13 @@ KAGGLE_DATASET_TOTAL_PAGES=500
 
 # Proxy Configuration
 PROXY_LIST=http://username:password@host:port,http://username:password@host:port
-PROXY_LOCATIONS=l1,l2
\ No newline at end of file
+PROXY_LOCATIONS=l1,l2
+
+# Wallet monitor
+WALLET_BALANCE_CHECK_INTERVAL_SECONDS=300
+WALLET_BALANCE_THRESHOLD_USDFC=1000000000000000000  # 1 USDFC (18 decimals)
```
beck-8: Hello, we would like to change the comment to. Because the current way of writing causes docker/docker-compose to not read env correctly
The rest of the configuration, I will also fix in another PR.
```
# 1 USDFC (18 decimals)
WALLET_BALANCE_THRESHOLD_USDFC=1000000000000000000
# 0.5 USDFC
WALLET_AUTO_FUND_AMOUNT_USDFC=500000000000000000
```

pali101:
Thanks for the context regarding filecoin-pin's existing auto-fund capabilities - I wasn't aware of that overlap.

I'll close this PR now.

PR #94, opened 2025-12-11
feat: separate web and backend infra

silent-cipher:
closes #85

PR #96, opened 2025-12-12
feat: use pnpm workspace for better dx

silent-cipher:
- use of pnpm workspace for better dx and project management
- simpler biome configs

This is follow up PR for #94


apps/backend/biome.json
```diff
@@ -0,0 +1,58 @@
+{
+  "$schema": "https://biomejs.dev/schemas/2.3.8/schema.json",
+  "root": false,
```
SgtPooki: not needed because `"extends": "//"`

biome.json
```diff
@@ -7,19 +7,23 @@
...
     "indentStyle": "space",
     "indentWidth": 2,
     "lineEnding": "lf",
-    "lineWidth": 120,
-    "attributePosition": "auto",
-    "bracketSameLine": false,
-    "bracketSpacing": true,
-    "useEditorconfig": true
+    "lineWidth": 120
+  },
```
SgtPooki: we should move this to the root Biome.json

biome.json
```diff
@@ -7,19 +7,23 @@
...
+  "assist": {
+    "enabled": true,
+    "actions": {
+      "source": {
+        "organizeImports": "on"
+      }
+    }
```
SgtPooki: and probably this

SgtPooki (changes requested):
other than setting private:true in the package.json and some biome.json cleanup (moving common rules to root biome.json) this is great.

SgtPooki (approved):

PR #98, opened 2025-12-12
fix: nest shutdown hooks and backend listening log

SgtPooki:


SgtPooki:
see https://github.com/FilOzone/dealbot/pull/100

PR #100, opened 2025-12-15
fix: nest shutdown hooks and backend listening log

SgtPooki:
replacement for https://github.com/FilOzone/dealbot/pull/98

### DIFF
--- a/apps/backend/src/main.ts
+++ b/apps/backend/src/main.ts
@@ -44,9 +44,7 @@ async function bootstrap() {
   const port = process.env.DEALBOT_PORT || 3000;
   const host = process.env.DEALBOT_HOST || "127.0.0.1";
   await app.listen(port, host);
-  logger.log(
-    `Dealbot backend is running on ${host}:${port}`,
-  );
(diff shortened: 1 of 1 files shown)


silent-cipher (approved):
Minor Suggestion - we could also consider adding explicit `SIGINT` / `SIGTERM` handlers (with `enableShutdownHooks`). Something like - 

```javascript
  process.on("SIGINT", async () => {
    console.log("Received SIGINT, shutting down gracefully...");
    await app.close();
    process.exit(0);
  });

  process.on("SIGTERM", async () => {
    console.log("Received SIGTERM, shutting down gracefully...");
    await app.close();
    process.exit(0);
  });
```

Not required for this PR, but could be a nice improvement for safer shutdowns.

SgtPooki:
@silent-cipher I may be missing something, i'm not intimately familiar with NestJS. How is explicit signals safer for shutdowns? If we add those listeners ourselves, it would duplicate what enableShutdownHooks is doing, wouldn't it?

`enableShutdownHooks` will also call module/service hooks for shutting down, so each piece of the app can ensure it "does the right thing" before shutdown of the entire app.

PR #115, opened 2026-01-06
fix: add http request and retrieval job timeouts

### LINKED ISSUE #113: bug: The retrieval task cannot be ended
https://dealbot.fwss.io/

It seems that the retrieval task at this time is abnormally stuck at 2025.12.30 04:00:00 (utc), causing subsequent tasks to be unable to proceed.

Image

Image

Image

SgtPooki:
# Prevent retrieval jobs from hanging

## Why
Retrieval tests could hang indefinitely (notably HTTP/2 requests without timeouts), which caused scheduler overlap and “Previous retrieval test still running” warnings.

## What changed
- We now force any existing retrieval jobs to abort when a new one starts.
- Added configurable timeout settings and wired them into the app config.
- HTTP/2 requests now have a dual-timeout strategy: fast-fail connection/headers timeout + total transfer timeout.
- HTTP/1.1 requests now use a configurable transfer timeout instead of a hardcoded 30s.
- Retrieval batches now respect a global deadline; we skip starting a batch unless there’s enough time for a full HTTP timeout.
- Added small timeout utilities and updated scheduling to enforce a global kill switch.

## Config
Add to `.env` (defaults shown):
```
CONNECT_TIMEOUT_MS=10000
HTTP_REQUEST_TIMEOUT_MS=600000
HTTP2_REQUEST_TIMEOUT_MS=600000
RETRIEVAL_TIMEOUT_BUFFER_MS=60000
```

## Added tests

* apps/backend/src/common/utils.spec.ts
* apps/backend/src/http-client/http-client.service.spec.ts
* apps/backend/src/retrieval/retrieval.service.spec.ts

## Notes

When we migrate to a job queue, a lot of this work will go away.. 

Fixes #113

### DIFF
--- a/apps/backend/src/config/app.config.ts
+++ b/apps/backend/src/config/app.config.ts
@@ -28,7 +28,31 @@ export const configValidationSchema = Joi.object({
 
   // Scheduling
   DEAL_INTERVAL_SECONDS: Joi.number().default(30),
-  RETRIEVAL_INTERVAL_SECONDS: Joi.number().default(60),
+  RETRIEVAL_INTERVAL_SECONDS: Joi.number()
+    .min(1)
+    .default(60)
+    .custom((value, helpers) => {
+      const root = helpers.state.ancestors[0] as {
+        RETRIEVAL_TIMEOUT_BUFFER_MS?: number;
+        HTTP_REQUEST_TIMEOUT_MS?: number;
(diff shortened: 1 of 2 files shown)


beck-8:
Looks great, a lot of improvements over the original search.
It turns out that there are too many aborts due to the 30-second limit. We will continue to observe the situation of the new version when it is run later.

PR #117, opened 2026-01-08
feat(backend): enhance CORS configuration with * support

beck-8:
enhance CORS with wildcard support


beck-8:
@SgtPooki Thanks for your guidance, please review again

beck-8:
I think I might actually understand you. It should meet the requirements now. Part of my knowledge was mixed up. It turns out that `*.xx.com` is not a standard.

Issue #69, opened 2025-12-02
Filecoin-Pin integration


rjan90:
## Description
Dealbot currently uses Synapse-SDK to create deals, but we should also leverage filecoin-pin for the IPN-indexing and IPFS-retrieval testing/metrics, so that we leverage our stack, instead of relying on custom implementations for these metrics.

SgtPooki:
we need to make sure that deal SUCCESS is dependent upon the filecoin-pin upload flow that validates IPNI + SP retrieval..

PR #124, opened 2026-01-15
feat: use filecoin-pin for IPNI validation

SgtPooki:
## Summary
Swapped Dealbot’s custom IPNI verification with filecoin-pin’s waitForIpniProviderResults, removing ~240 LOC of bespoke HTTP/validation logic and aligning behavior with the filecoin-pin CLI/GitHub Action.

## What changed
- Added dependency: filecoin-pin (^0.14.0)
- `monitorAndVerifyIPNI()` now delegates to `waitForIpniProviderResults`
- Added `buildExpectedProviderInfo()` to map Dealbot StorageProvider → Synapse ProviderInfo
- Safer null handling for `deal.storageProvider` in `startIpniMonitoring()`
- Deleted old IPNI verification/query helpers (and unused imports)

## Behavior note
Verification is now root CID only (no per-block CID checks), matching filecoin-pin’s current behavior. Counts now reflect that (total = 1, verified = 0/1). Curio, as far as I can tell, does not announce individual block CIDs inside a rootCID.

follow-up for child block checks should be handled by filecoin-pin. tracking issue at https://github.com/filecoin-project/filecoin-pin/issues/303

## Why
- One shared source of truth for IPNI verification
- Less code to maintain in Dealbot
- More consistent results across tooling

## Related

- #69

### DIFF
--- a/apps/backend/src/deal-addons/strategies/ipni.strategy.ts
+++ b/apps/backend/src/deal-addons/strategies/ipni.strategy.ts
@@ -204,10 +204,14 @@ export class IpniAddonStrategy implements IDealAddon<IpniMetadata> {
       const serviceUrl = deal.storageProvider.serviceUrl;
       const pdpServer = new PDPServer(null, serviceUrl);
 
+      const rootCID = deal.metadata[this.name]?.rootCID ?? "";
+      const blockCIDs = deal.metadata[this.name]?.blockCIDs ?? [];
+
       const result = await this.monitorAndVerifyIPNI(
         pdpServer,
-        deal.pieceCid,
-        deal.metadata[this.name]?.rootCID ?? "",
+        deal,
+        blockCIDs.map((cid) => CID.parse(cid)),
+        rootCID,
         deal.storageProvider,
         this.POLLING_TIMEOUT_MS,
         this.IPNI_LOOKUP_TIMEOUT_MS,
@@ -234,7 +238,8 @@ export class IpniAddonStrategy implements IDealAddon<IpniMetadata> {
 
   async monitorAndVerifyIPNI(
     pdpServer: PDPServer,
-    pieceCid: string,
+    deal: Deal,
+    blockCIDs: CID[],
     rootCID: string,
     storageProvider: StorageProvider,
     statusTimeoutMs: number,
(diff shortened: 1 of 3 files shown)


apps/backend/src/deal-addons/strategies/ipni.strategy.ts
```diff
@@ -241,8 +269,39 @@ export class IpniAddonStrategy implements IDealAddon<IpniMetadata> {
...
     }
 
-    // Always verify CIDs via filecoinpin.contact - this is the terminal state
-    const ipniResult = await this.verifyIPNIAdvertisement(rootCID, blockCIDs, expectedMultiaddr, ipniTimeoutMs);
+    const rootCidObj = CID.parse(rootCID);
+    const maxAttempts = Math.ceil(ipniTimeoutMs / 5000 / 2);
+
+    this.logger.log(`Verifying rootCID in IPNI: ${rootCID.slice(0, 12)}...`);
+
+    // NOTE: filecoin-pin does not currently validate that all blocks are advertised on IPNI.
```
silent-cipher: Do we have a tracking issue for adding support to validate that all block cids are advertised on IPNI?

apps/backend/src/deal-addons/strategies/ipni.strategy.ts
```diff
@@ -22,15 +23,37 @@ import type {
...
+    active: storageProvider.isActive,
+    products: {
+      PDP: {
+        type: "PDP",
+        isActive: true,
+        capabilities: {},
+        data: {
+          serviceURL: storageProvider.serviceUrl,
+          // we don't need the other fields for IPNI verification
+        } as any,
```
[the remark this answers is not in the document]
SgtPooki: this is safe, provider is only used for its service url: https://github.com/filecoin-project/filecoin-pin/blob/f22a6c64f1737d3e95feff2332f00a95364d02b1/src/core/utils/validate-ipni-advertisement.ts#L335-L381

SgtPooki:
> One concern I have is about removing validation of block cids and relying solely on root cid verification. Could we consider a hybrid approach instead? For example, using filecoin-pin to validate the root cid, while keeping a custom validation mechanism for block cids.

I have a tracking issue in filecoin-pin to handle block CID verification.. this should be handled in there so we're not duplicating code.

I can work on that before we merge this though.

apps/backend/src/deal-addons/strategies/ipni.strategy.ts
SgtPooki: we do now! https://github.com/filecoin-project/filecoin-pin/issues/303

PR #125, opened 2026-01-16
fix(web): display 'Unknown' for curio version when SP version check fails

### LINKED ISSUE #99: ui consistency: SP curio version
We should display curio version as `unknown` or `not-reported` if `sp/version` check fails:

Image

virajbhartiya:
## Summary                                                        
- Display "Curio Version: Unknown" when SP version check fails instead of hiding the version section entirely                     
- Uses muted styling to indicate the version could not be retrieved                                                          

Closes #99

### DIFF
--- a/apps/web/src/components/Home/StorageProviders/ProviderCard/ProviderCardHeader.tsx
+++ b/apps/web/src/components/Home/StorageProviders/ProviderCard/ProviderCardHeader.tsx
@@ -8,7 +8,6 @@ interface ProviderCardHeaderProps {
   provider: Provider;
   version?: string;
   versionLoading: boolean;
-  versionError: string | null;
   hasMetrics: boolean;
   health: {
     status: string;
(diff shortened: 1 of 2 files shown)


apps/web/src/components/Home/StorageProviders/ProviderCard/ProviderCardHeader.tsx
SgtPooki:
```suggestion
      {versionLoading 
        ? (<Skeleton className="h-5 w-full" />) 
        : (
			<div className="text-sm flex justify-between items-center gap-2 mb-0">
			<p className="text-sm text-muted-foreground">Curio Version:</p>
			<span className="font-medium">{version ?? 'Unknown'}</span>
			</div>
)}
```

SgtPooki: formatting may be off, but we don't need a whole new HTML and further nest ternaries. nested ternaries are the devil.

SgtPooki (approved):
lgtm

PR #127, opened 2026-01-16
fix: create random data when others fail

SgtPooki:
# Add Random Dataset Generation Fallback

## Current State
Deal creation fails if both Kaggle datasets and local datasets are unavailable.

## This PR
Adds random dataset generation as a third fallback tier, currently just to unblock some errors seen on dealbot deployed to new infra:
- Generates unique random data (10 KiB, 10 MB, or 100 MB) when other sources fail
- Automatically cleans up generated files after upload
- Configurable via `RANDOM_DATASET_SIZES` environment variable
- Uses streaming to minimize memory usage (~50% reduction)

## Related Issues
- Related https://github.com/FilOzone/infra/issues/28
- Related https://github.com/FilOzone/infra/issues/29
- Supports testing dApp sizes and throughput scenarios as discussed by @timfong888

## NOTE

eventually, we should probably use ONLY this method..

### DIFF
--- a/apps/backend/src/common/stream-utils.spec.ts
+++ b/apps/backend/src/common/stream-utils.spec.ts
@@ -1,6 +1,6 @@
+import { Writable } from "node:stream";
 import * as fs from "fs";
 import * as path from "path";
-import { Writable } from "node:stream";
 import { afterEach, beforeEach, describe, expect, it } from "vitest";
 import { writeWithBackpressure } from "./stream-utils.js";
 


apps/backend/src/dataSource/dataSource.service.ts
```diff
@@ -110,4 +111,135 @@ export class DataSourceService {
...
+      writeStream.write(prefixBuffer);
+
+      // Write random data in chunks
+      let remainingBytes = randomDataSize;
+      while (remainingBytes > 0) {
+        const currentChunkSize = Math.min(chunkSize, remainingBytes);
+        const chunk = crypto.randomBytes(currentChunkSize);
+        writeStream.write(chunk);
+        remainingBytes -= currentChunkSize;
+      }
```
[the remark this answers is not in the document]
SgtPooki: fix incoming.

apps/backend/src/common/stream-utils.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+        const buffer = Buffer.alloc(256); // 256 bytes > 32 bytes highWaterMark
+        await writeWithBackpressure(writeStream, buffer);
+      }
+
+      writeStream.end();
+      await new Promise<void>((resolve) => writeStream.on("finish", () => resolve()));
+
+      // Should have triggered drain multiple times (since we're doing 5 in parallel, at least 4 drains are expected)
+      // see https://nodejs.org/api/stream.html#event-drain
+      expect(drainCount).toBeGreaterThanOrEqual(4);
```
silent-cipher: This comment seems to be misleading -

```suggestion
      // Should have triggered drain multiple times (since we're writing 5 chunks that each exceed highWaterMark, we expect backpressure to trigger each time)
      // see https://nodejs.org/api/stream.html#event-drain
      expect(drainCount).toBe(5);
```

SgtPooki: yep, good call

PR #128, opened 2026-01-16
fix: release to prod PRs can be added to filoz project

SgtPooki:
see https://github.com/FilOzone/dealbot/pull/121#issuecomment-3761678327 for details


rjan90:
Just an FYI: I approved a `dealbot_release_please_pr_PAT` PAT on FilOzzy, which I suspect was created due to this PR?

PR #131, opened 2026-01-20
fix(backend): add optional dealbot dataset versioning

### LINKED ISSUE #102: fix: hardcode a version number for dataSet metadata
if we need to create new dataSets for whatever reason, we should be able to do so without creating a new wallet. we can do this by having a hardcoded version number in the src code that we set in the dataset metadata.

Something like `dealbot-v1` or similar

silent-cipher:
Introduces a new optional env variable, `DEALBOT_DATASET_VERSION`.
If provided, its value is added to the sdk dataset metadata as `version`.

fixes #102

### DIFF
--- a/apps/backend/src/deal/deal.service.spec.ts
+++ b/apps/backend/src/deal/deal.service.spec.ts
@@ -258,8 +258,8 @@ describe("DealService", () => {
         expect(mockSynapseInstance.createStorage).toHaveBeenCalledWith({
           providerAddress: "0xProvider",
           metadata: {
-            version: "dealbot-v2",
             customKey: "customValue",
+            dealbotDataSetVersion: "dealbot-v2",
           },
         });
       });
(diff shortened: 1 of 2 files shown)


apps/backend/src/config/app.config.ts
```diff
@@ -181,6 +183,7 @@ export function loadConfig(): IConfig {
       useOnlyApprovedProviders: process.env.USE_ONLY_APPROVED_PROVIDERS === "true",
       enableCDNTesting: process.env.ENABLE_CDN_TESTING === "true",
       enableIpniTesting: process.env.ENABLE_IPNI_TESTING === "true",
+      dealbotDatasetVersion: process.env.DEALBOT_DATASET_VERSION,
```
SgtPooki: naming standardization from prior conversations with Rod: Use two-word form: DataSet or data set or data-set, not "Dataset".

```suggestion
      dealbotDataSetVersion: process.env.DEALBOT_DATASET_VERSION,
```

PR #140, opened 2026-01-21
fix(backend): negative avg time to retrieve metrics

### LINKED ISSUE #63: IPNI indexing: Average time to retrieve metrics is negative
The IPNI Indexing task `average time to retrieve` is negative across the board of SPs. So I suspect there is some wrong calculation going on:

Image

silent-cipher:
This PR includes - 

1. We do not default`uploadEndTime` to `now`. If it is missing, we simply skip the duration calculation (metrics are meaningless without a valid start time).
2. Added `Math.max(0, ...)` to ensure that even with slight clock skews, we never record negative durations.
3. If `retrievedAt` is missing from sdk, fallback to time when dealbot catches status change to `retrieved`

Closes #63

### DIFF
--- a/apps/web/src/components/Home/StorageProviders/ProviderCard/ProviderCardDetailedStats.tsx
+++ b/apps/web/src/components/Home/StorageProviders/ProviderCard/ProviderCardDetailedStats.tsx
@@ -129,7 +129,6 @@ function ProviderCardDetailedStats({
               <div className="pl-2 space-y-1">
                 <LatencyMetricRow label="Time to Index" value={allTimeMetrics.avgIpniTimeToIndexMs} />
                 <LatencyMetricRow label="Time to Advertise" value={allTimeMetrics.avgIpniTimeToAdvertiseMs} />
-                <LatencyMetricRow label="Time to Retrieve Request" value={allTimeMetrics.avgIpniTimeToRetrieveMs} />
                 <LatencyMetricRow label="Time to Verify" value={allTimeMetrics.avgIpniTimeToVerifyMs} />
               </div>
             </div>


apps/backend/src/deal-addons/strategies/ipni.strategy.ts
```diff
@@ -582,30 +588,50 @@ export class IpniAddonStrategy implements IDealAddon<IpniMetadata> {
       deal.ipniStatus = IpniStatus.FAILED;
     }
 
+    // Helper function to calculate duration in milliseconds
+    const calculateDuration = (eventTime: Date): number => {
+      const duration = eventTime.getTime() - uploadEndTime.getTime();
+      return Math.max(0, duration);
+    };
```
SgtPooki: should we instead return null? since this value is used to calculate avg, if we don't have a valid duration, we need to ignore it.

also, we should probably update the query to get the value for the frontend (or the metrics job.. I forget where it is offhand) to filter out any null or <= 0 issues. duration of <= 0 should be impossible right?

SgtPooki (commented):
some comments about consistency and goals of the metric..

apps/backend/src/deal-addons/strategies/ipni.strategy.ts
silent-cipher: Yes, returning `null` makes more sense here. I originally had that, but ran into a TypeScript error due to a mismatch between the deal entity schema and the class type definitions (which should probably be fixed). To move forward, I ended up ignoring it - not the best call on my part. I'll add this.

The existing queries already filter out `null` values, and with the introduction of `Math.max`, a `duration <= 0` should no longer be possible.

One question: should we also explicitly return `null` if the computed `duration` evaluates to a negative value, just to be defensive?

silent-cipher: Marking it resolved. Incorporated your suggested changes.

PR #150, opened 2026-01-22
fix: better scheduling after restarts

### LINKED ISSUE #119: Change task scheduling to use startup-relative intervals instead of wall-clock time
Pros:
- Some SPs deploy the test network and main network in the same computer room, which can separate different network environments and avoid a backlog of uplink and downlink bandwidth at the same time.
- Multiple dealbot instances can initiate tasks at different times, same as above
- Avoid calling glif rpc at the same time to prevent exceeding the limit (if there are instances of mainnet and testnet in a network environment)

Cons:
- ?

SgtPooki:
* Schedule deal/retrieval runs from last created_at + interval; use start offsets only for fresh DBs.
* Metrics jobs now anchor to created_at/refreshed_at with initial offsets; weekly cleanup stays cron‑based.
* Documented start offsets as “first run only” in env var docs.

fixes #119

### DIFF
--- a/apps/backend/src/deal/deal.service.spec.ts
+++ b/apps/backend/src/deal/deal.service.spec.ts
@@ -36,6 +36,7 @@ describe("DealService", () => {
   const mockDealRepository = {
     create: vi.fn(),
     save: vi.fn(),
+    createQueryBuilder: vi.fn(),
   };
 
   const mockStorageProviderRepository = {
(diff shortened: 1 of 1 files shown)


SgtPooki:
i'm going to enable a default-off pg-boss implementation that should solve #119 and help us better manage scheduling between restarts/downtime

Issue #160, opened 2026-01-27
Document "check" events and metrics


BigLep:
## Done Criteria
There is a public document that lists
- The set of "events" that occur within the various "dealbot" checks (e.g., "Data Storage", "Retrieval") including the event definition and source of truth
- Metrics that are computed from these checks

## Why Important
1. Drives clarity for engineering team on what to implement
2. Gives a canonical resource for the dealbot dashboard to link to so users can know what a metric means and how it's computer
 
## User/Customer
1. Maintainers
2. Dashboard consumers who have questions about what metrics mean

## Notes
1. Markdown file in the repo is fine.
2. Explorative drafting has been happening in https://www.notion.so/filecoindev/dealbot-metric-scratchpad-2f1dc41950c1808b94e1fe85babc4f00

Issue #158, opened 2026-01-26
GA "Data Storage" check implementation and documentation


BigLep:
## Done Criteria
1. There is public documentation for what dealbot does for its "Data Storage" check (e.g., how many datasets it creates per SP, what it asserts every run)
2. There is implementation that matches this documentation

## Why Important
We've been talking/writing for a bit now about what Dealbot should be checking.  We need to get that formalized in public documentation (that the dealbot UI can link to) so dealbot dashboard readers can understand the system and so engineers have a clear target of what to implement.
 
## User/Customer
1. Dealbot maintainers
2. Dealbot dashboard consumers

## Notes
1. Markdown in this repo is fine.
2. There is likely content from https://linear.app/filoz/project/robust-dealbot-user-provingexperience-fa6e2ea06a06/overview we can draw from.
3. Documentation should link to source code as the place for someone to see the source of truth

Issue #123, opened 2026-01-15
Determine GA scope


BigLep:
## Done Criteria
Engineering understands what functionality is required for GA and has a set of issues for the remaining work.  

The biggest open area is what are the "SP Acceptance Service Level Indicators (SLIs) and Service Level Objectives (SLOs)" that must be hit by an AP to be added, removed, and readded to the FWSS Approved List.

Second is what are the additional metrics we want to display about each SP.

## Notes
1. Discussion is happening in https://linear.app/filoz/project/robust-dealbot-user-provingexperience-fa6e2ea06a06/overview

BigLep:
This work is close to done with:
- UI scope defined
   - https://github.com/FilOzone/dealbot/issues/158
   - https://github.com/FilOzone/dealbot/issues/159
- Acceptance thresholds: https://linear.app/filoz/document/public-definition-of-thresholds-for-dealbot-ccf29ae32583

It will finish as we do the documentation items, which will for clearing up scope ambiguity and requirements: https://github.com/FilOzone/dealbot/pull/164

PR #164, opened 2026-01-28
docs: define deals, retrievals, and metrics

SgtPooki:
This PR attempts to handle the documentation parts of #158, #159, #160.

This is all in a single PR to make the entire scope of documentation changes easier to review.

### DIFF
--- a/docs/checks/data-storage.md
+++ b/docs/checks/data-storage.md
@@ -1,6 +1,6 @@
 # Data Storage Check
 
-This document is the **source of truth** for how dealbot's Data Storage check works. Items marked **TBD** are not yet implemented; code changes will follow.
+This document is the **source of truth** for how dealbot's Data Storage check works. (Items marked **TBD** are not yet implemented; code changes will follow.)
 
 Source code links throughout this document point to the current implementation.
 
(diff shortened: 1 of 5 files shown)


docs/checks/data-storage.md
```diff
@@ -0,0 +1,273 @@
...
+
+### 3. Upload to Each SP
+
+For each **testing SP**, dealbot:
+
+1. **Creates a dataset on-chain** via the Synapse SDK (`synapse.createStorageContext()`), one dataset per SP. Dataset creation is idempotent (if a dataset with the same metadata already exists for the SP, it is reused).
+2. **Uploads the CAR file** to the SP. Three callbacks track progress:
+   - `onUploadComplete` — SP confirms receipt. Records the piece CID, upload latency, and throughput.
+   - `onPieceAdded` — piece submission is recorded (transaction hash available). Indexing in the SP/Curio DB should be complete.
+   - `onPieceConfirmed` — piece is confirmed on-chain. Records chain latency.
```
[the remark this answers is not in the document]
BigLep: Maybe there is some valid things here @SgtPooki ?

docs/checks/data-storage.md
```diff
@@ -0,0 +1,273 @@
...
+**Failure** occurs if any step fails or the deal exceeds its max allowed time. There are no timing-based quality assertions. Operational timeouts exist to prevent jobs from running indefinitely, but they are not quality assertions. A per-deal max time limit that fails the deal if exceeded is **TBD**.
+
+## What Gets Asserted
+
+Each deal asserts the following for every SP:
+
+| # | Assertion | How It's Checked | Implemented? |
+|---|-----------|-----------------|:---:|
+| 1 | SP accepts data upload | Upload completes without error; piece CID is returned | Yes |
+| 2 | Piece submission recorded on-chain | `onPieceAdded` callback fires with a transaction hash | Yes |
```
BigLep: Where are the callbacks that you're referring to?

docs/checks/events-and-metrics.md
```diff
@@ -0,0 +1,86 @@
...
+## Event Model
+
+Events are grouped by check type. Each event includes a short definition, its implementation status, and a link to the expected source of truth.
+
+### Data Storage Events
+
+| Event | Definition | Status | Source of truth |
+|------|------------|:------:|-----------------|
+| `uploadToSpStart` | Dealbot begins attempting to send a piece to an SP. | **TBD** | [`deal.service.ts`](../../apps/backend/src/deal/deal.service.ts) |
+| `uploadToSpEnd` | Upload finishes; piece CID is known; ingest metrics recorded. | Yes | [`deal.service.ts`](../../apps/backend/src/deal/deal.service.ts) (`handleUploadComplete`) |
```
BigLep:
```suggestion
| `uploadToSpEnd` | Upload finishes by receiving a 2xx response from the PDP server; piece CID is known; ingest metrics recorded. | Yes | [`deal.service.ts`](../../apps/backend/src/deal/deal.service.ts) (`handleUploadComplete`) |
```

In general, be more specific where we can so it's clear.

BigLep (commented):
Thanks for putting this together Russell.  I think this will be a useful artifact going forward, and I think it's useful right now for making sure we're fully aligned on what https://github.com/FilOzone/dealbot/issues/158 and https://github.com/FilOzone/dealbot/issues/159 should implement.  I think it was smart to start with it.

I have read through everything, but because of time I didn't comment on each thing I saw.  I have a few inline comments and some more general things.  If I had more time tonight, I would have just created a PR on top of yours to point out some other suggestions or changes.  I think in terms of next steps it would be great if you can review/incorporate the feedback I have provided so far.  I'll then checkout the change and as I read through it again, make any other changes in a PR on top so it's easy for you to review/accept/reject my items.  Sound good?

Generall items

- I think we should write this document assuming the flow for when https://github.com/FilOzone/dealbot/issues/158 and https://github.com/FilOzone/dealbot/issues/159 are done (since they will be completed soon).  It's fine/good to denote the parts that aren't complete as TBD or TBI (to be implemented).  For example, lets write it assuming the IPNI advertising and verification happens as part of the sequential "data storage" flow  
- What retries are we doing?
- Rather than stating line number in text can we just `#123` in the URLs (more convenient and less text presented to user)
- Maybe this is a losing battle, but rather than use the word "deal" throughout the document, do we use "piece" since that has a definition throughout FWSS?
- Some parts in each document seem duplicative?  Maybe ask AI for ideas to reduce redundancy?
- In particular, there seems to be some duplication of the metric definitions across the files?  Either reference events-and-metrics.md as much as possible or inline that content into the other two files?
- Not critical, but maybe do mermaid diagrams?

BigLep:
I'll approve this once https://github.com/FilOzone/dealbot/pull/228 is merged.

PR #166, opened 2026-01-28
fix: import getPieceStatus logic

SgtPooki:
# Remove `getPieceStatus` from synapse-sdk and improve IPNI monitoring

## Summary

Prepares for upcoming synapse-sdk changes while fixing IPNI monitoring stability issues:

1. **Implement `getPieceStatus` locally** - Moved from synapse-sdk to our own `HttpClientService`
2. **Remove unstable `retrieved` status** - Deprecated `sp_received_retrieve_request` (non-deterministic, depends on external requests)
3. **Exit monitoring on `advertised`** - `monitorPieceStatus` now returns as soon as piece is advertised instead of waiting for retrieved
4. **Remove 30s artificial delay** - No more hard-coded wait after monitoring completes
5. **Fix timing metrics** - Use `deal.uploadEndTime` instead of `Date.now()` for accurate metric calculations

## Key Changes

- Added `getPieceStatus()` to `HttpClientService` with proper validation
- Removed `retrieved`, `retrievedAt`, and `ipniTimeToRetrieveMs` tracking
- IPNI verification now starts immediately after piece is advertised (faster, more reliable)
- Method signatures changed: `pdpServer: PDPServer` → `serviceURL: string`

## Impact

- Faster feedback (no 30s wait)
- More reliable (doesn't depend on external retrieve requests)
- Accurate metrics (calculated from upload completion, not monitoring start)

## Related

#158 
#159
#160

### DIFF
--- a/apps/backend/src/deal-addons/strategies/ipni.types.ts
+++ b/apps/backend/src/deal-addons/strategies/ipni.types.ts
@@ -20,7 +20,9 @@ const pieceStatusResponseSchema = Joi.object({
   status: Joi.string().required(),
   indexed: Joi.boolean().required(),
   advertised: Joi.boolean().required(),
-}).required();
+})
+  .unknown(true)
+  .required();
 
 /**
  * Type guard for PieceStatusResponse


silent-cipher (approved):
Looks good to me!
Flagging that there are some Copilot comments above to follow up on.

PR #228, opened 2026-02-05
docs: improve checks documentation - fix typos, links, and grammar

BigLep:
This PR has the set of things that I would want to change / improve or discuss improving on https://github.com/FilOzone/dealbot/pull/164

PR #240, opened 2026-02-09
feat: split worker and api

SgtPooki:
**Worker Mode + Pg-Boss Tuning**

**Summary**
- Add `DEALBOT_RUN_MODE` (`api|worker|both`, default `both`) to split API/scheduler from workers or run all-in-one.
- Introduce worker-only boot path with metrics-only HTTP listener.
- Add `DEALBOT_PGBOSS_SCHEDULER_ENABLED` to disable enqueue loop in worker pods.
- Add `JOB_WORKER_POLL_SECONDS` (min 5s, default 60s) and pass to pg-boss `subscribe()`.
- Default concurrency to 10/10 and update docs.

**Why**
- Enable scalable worker replicas while keeping a single API/scheduler pod.
- Reduce API pod overhead and allow independent tuning of worker polling and concurrency.

**Key Files**
- `apps/backend/src/main.ts`
- `apps/backend/src/worker.module.ts`
- `apps/backend/src/metrics/metrics-worker.module.ts`
- `apps/backend/src/jobs/jobs.service.ts`
- `apps/backend/src/config/app.config.ts`
- `docs/environment-variables.md`

**Notes**
- Worker mode exposes `/metrics` on `DEALBOT_METRICS_PORT`.
- API mode keeps `/metrics` on `DEALBOT_PORT`.

---

**NOTE:** this is needed because job queue is growing larger than they're being processed. 

image

We may also need to cap runtime for deal jobs because some are lasting 5 minutes which slows the whole queue down..

Some maths:

we currently need to do **72 deals/hour** (18 providers × 4 deals/provider/hour).

If each job is 5 minutes, each concurrent slot can do **12 jobs/hour**.  
Minimum concurrency to hit 72/hour:

```
18 * 4 * 5min / 60min = 6 slots 
```

So:
- If concurrency ≥ 6 → max is **72/hour** (limited by your per‑provider rate).
- If concurrency = C → max is **min(72, 12*C)** deals/hour.

where 

```
concurrency = worker_replicas * teamSize
```

- `worker_replicas` = number of worker pods
- `teamSize` = `DEAL_MAX_CONCURRENCY` (per pod)

So if we run 3 worker pods with `DEAL_MAX_CONCURRENCY=4`, our total concurrency is **12**.

Also note per‑SP locks can reduce effective concurrency. (retrievals block deals for an SP and retrievals are typically running in less than a minute)

### DIFF
--- a/apps/backend/src/config/app.config.ts
+++ b/apps/backend/src/config/app.config.ts
@@ -6,7 +6,7 @@ import type { Network } from "../common/types.js";
 export const configValidationSchema = Joi.object({
   // Application
   NODE_ENV: Joi.string().valid("development", "production", "test").default("development"),
-  DEALBOT_RUN_MODE: Joi.string().valid("api", "worker", "both").default("both"),
+  DEALBOT_RUN_MODE: Joi.string().lowercase().valid("api", "worker", "both").default("both"),
   DEALBOT_PORT: Joi.number().default(3000),
   DEALBOT_HOST: Joi.string().default("127.0.0.1"),
   DEALBOT_METRICS_PORT: Joi.number().default(9090),
--- a/apps/backend/src/deal/deal.module.ts
+++ b/apps/backend/src/deal/deal.module.ts
@@ -6,7 +6,6 @@ import { Retrieval } from "../database/entities/retrieval.entity.js";
 import { StorageProvider } from "../database/entities/storage-provider.entity.js";
 import { DataSourceModule } from "../dataSource/dataSource.module.js";
 import { DealAddonsModule } from "../deal-addons/deal-addons.module.js";
-import { MetricsModule } from "../metrics/metrics.module.js";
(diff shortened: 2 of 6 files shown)


BigLep:
A few things @SgtPooki :
1. It makes sense why we need/want this.
2. This is a pretty big/important architectural change.  I think it would be great to get other eyes on it, even if you still need to merge for now to unblock yourself.  So basically, I think it would be good to request reviewers (e.g., @silent-cipher or someone from Space Meridian) and not have it in the done status so we give them a chance to look at it, even if it's merged.
3. "We may also need to cap runtime for deal jobs because some are lasting 5 minutes which slows the whole queue down.." Agreed.  I think the step that would be good here is to finish off #158 (where finish off means it matches the "spec" from https://github.com/FilOzone/dealbot/pull/228 ).  Doing that would involve setting proper job timeouts.  Basically I think it's fine to make sure our checks are implemented correctly and then we can make sure they run at the desired frequency.  If we need to pause dealbot untitl then, I think it's ok.

PR #247, opened 2026-02-10
fix: use single pgboss queue to enforce per SP lock

SgtPooki:
## Summary
- Replace the session-scoped advisory lock with pg-boss `singletonKey` on a single `sp.work` queue to prevent concurrent deal/retrieval jobs for the same SP across all workers.
- Simplify pg-boss configuration to a single concurrency knob (`PG_BOSS_LOCAL_CONCURRENCY`) and remove unused/obsolete settings from docs and `.env.example`.
- Add runbook/docs guidance on capacity/backlog and clarify job status vs check outcome semantics.

## Problem
Staging showed spikes (e.g., 72 deals in flight) despite a “per‑SP” lock. The advisory lock was session‑scoped, so with pooled connections and per‑worker concurrency, multiple jobs could run against the same SP on a single worker. With multiple workers, this allowed many concurrent deal/retrieval jobs for the same SP.

## Fix
- Use pg‑boss `singletonKey=spAddress` on the `sp.work` queue to enforce a **global per‑SP exclusion** across all workers.
- Remove the unused `job_mutex` approach and stale‑lock config; per‑SP uniqueness is handled by pg‑boss.
- Consolidate pg‑boss concurrency to `PG_BOSS_LOCAL_CONCURRENCY` and document it alongside job scheduling parameters.

## Notes
- Lock remains **global per SP** (deal + retrieval share the same queue with `singletonKey`).
- Backfill is bounded by `JOB_CATCHUP_MAX_ENQUEUE`; no extra catch‑up jitter/spread knobs.
- See `docs/runbooks/jobs.md` and `docs/jobs.md` for capacity formulas, scheduling behavior, and job‑status vs check‑result semantics.


apps/backend/src/jobs/jobs.service.ts
```diff
@@ -254,15 +265,20 @@ export class JobsService implements OnModuleInit, OnApplicationShutdown {
...
       this.logMaintenanceSkip(`deal job for ${spAddress}`, maintenance.window?.label);
       return;
     }
 
-    const acquired = await this.tryAcquireSpLock(spAddress);
+    const jobId = typeof job.id === "string" ? job.id : randomUUID();
+    if (!job.id) {
+      this.logger.warn(`Deal job missing id; generated fallback id ${jobId} for ${spAddress}`);
+    }
+    const acquired = await this.tryAcquireSpMutex("deal", spAddress, jobId);
```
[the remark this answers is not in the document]
SgtPooki: they are UUIDs. see https://github.com/timgit/pg-boss/blob/master/docs/api/jobs.md

docs/jobs.md
BigLep: It would be nice to see a mapping between job_type and the source code that handles that job type.

BigLep: Can we get rid of non-pgboss scheduling to simplify?

docs/jobs.md
```diff
@@ -0,0 +1,216 @@
...
+4. **SP status changes (example)**: When `loadProviders()` syncs updated provider status into `storage_providers`, the next scheduler tick re-evaluates active/approved providers. If a provider is no longer active or approved, its deal/retrieval schedules are deleted.
+
+Note: In pg-boss mode, provider sync currently happens at startup (and whenever `loadProviders()` is called). There is no periodic refresh loop, so new/changed providers require a refresh (restart or an explicit call to `loadProviders()`).
+
+## How `next_run_at` Changes
+
+`next_run_at` is controlled entirely by the scheduler loop:
+
+- **Initial value**: set by `upsertSchedule()` when schedules are created, using `now + JOB_SCHEDULE_PHASE_SECONDS`.
+- **On each tick**: The scheduler finds rows where `next_run_at <= now`, computes how many runs are due based on `interval_seconds`, enqueues up to `JOB_CATCHUP_MAX_ENQUEUE` runs per schedule row, and on successful enqueue advances `next_run_at` by `successCount * interval_seconds` while updating `last_run_at`.
```
BigLep: Why do we need catchup behavior?  This seems like complexity that we don't need.

BigLep: `successCount * interval_seconds` seems odd to me

So lets say we have catchup behavior.
We decided we should catchup 4 jobs.
We successfully enquee 4 jobs.
Assuming interval sectons is "15 minutes" that means we're going to set next_run_at for 4x15m = 1hour from now.  
1 hour from now we'll then enque one job if IIUC.
So it doesn't actually seem like we're doing any catchup.   

Per above I would just remove catchup functionality to keep it simple, or if we do want/need cathcup functionality, I think the algorithm needs to adjust (but maybe I'm misunderstanding it).

docs/jobs.md
```diff
@@ -0,0 +1,216 @@
...
+  BossJobs --> DB
+
+  Workers --> BossJobs
+  Workers --> MutexTable
+  Workers --> DB
+```
+
+## Parallelism and Limits
+
+- **Queue concurrency**: Deal queue concurrency is `DEAL_MAX_CONCURRENCY` (pg-boss [`batchSize`](https://github.com/timgit/pg-boss/blob/master/docs/api/workers.md)), retrieval queue concurrency is `RETRIEVAL_MAX_CONCURRENCY`, and metrics/cleanup are fixed at `batchSize=1`. These limits apply per instance; total concurrency scales with the number of worker pods.
```
BigLep: We have some terminology we should define here: "instance" and "worker pods".  I believe I know what these mean, but I think it would be good to define these.

Basically I assume by "instance" we mean each "nodejs process with DEALBOT_RUN_MODE=worker".   Basically this is a "dealbot worker proess"?

Using the term "worker pods" I think is bleading in k8s terminology?  Can we instead say "dealbot worker"?

Another terminology thing we need to be clear about here is a "dealbot worker" vs. a "pg-boss worker".  IIUC, there are many "pg-boss workers per dealbot worker".

docs/jobs.md
```diff
@@ -0,0 +1,216 @@
...
+- **Per-SP mutex**: `job_mutex` ensures only one job runs per SP at a time across both deal and retrieval queues. If a lock is held, the job is re-queued after `JOB_LOCK_RETRY_SECONDS`.
+
+Source: [`jobs.service.ts`](../apps/backend/src/jobs/jobs.service.ts), [`job-schedule.repository.ts`](../apps/backend/src/jobs/repositories/job-schedule.repository.ts).
+
+## Backpressure and Catch-Up
+
+- **Catch-up cap**: `JOB_CATCHUP_MAX_ENQUEUE` limits how many runs are enqueued per schedule row per tick.
+- **Staggering**: `JOB_CATCHUP_SPREAD_HOURS` spreads catch-up jobs, and `JOB_ENQUEUE_JITTER_SECONDS` adds random delay.
```
BigLep: What does this buy us?  If we just set a max number of pg-boss workers per dealbot-worker, why do we need to stagger?

docs/jobs.md
```diff
@@ -0,0 +1,216 @@
...
+```
+
+This staggers schedules by 20 minutes and randomizes starts within 5 minutes.
+
+## Is There a "Job Run" Record?
+
+- Dealbot does **not** store a first-class "job run" entity.
+- `job_schedule_state.last_run_at` is updated when schedules are advanced (enqueue time), not when handlers finish.
+- pg-boss itself stores each queued job in `pgboss.job` with states like `created`, `active`, and `retry` (we query this table for queue metrics).
+- Execution metrics are exported via Prometheus (`jobs_started_total`, `jobs_completed_total`, `job_duration_seconds`).
```
BigLep: I'm not sure where else this belongs, but it seems a bit out of place here.

docs/jobs.md
```diff
@@ -0,0 +1,216 @@
...
+
+Source: [`jobs.service.ts`](../apps/backend/src/jobs/jobs.service.ts), [`job-schedule.repository.ts`](../apps/backend/src/jobs/repositories/job-schedule.repository.ts), [`metrics-prometheus.module.ts`](../apps/backend/src/metrics-prometheus/metrics-prometheus.module.ts).
+
+## Critical Environment Variables
+
+See the "Jobs (pg-boss)" section in [`docs/environment-variables.md`](./environment-variables.md) for full definitions. The most important knobs are:
+
+- `DEALBOT_JOBS_MODE`
```
BigLep: Bonus: It would be nice if all of these were hyperlinks into environment-variables (like we're doing with the event and metric names).

docs/jobs.md
```diff
@@ -0,0 +1,216 @@
...
+Job types map to pg-boss queue names in [`JobsService.mapJobName`](../apps/backend/src/jobs/jobs.service.ts):
+
+- `deal` -> `deal.run`
+- `retrieval` -> `retrieval.run`
+- `metrics` -> `metrics.run`
+- `metrics_cleanup` -> `metrics.cleanup`
+
+Jobs are enqueued via [`boss.send(name, data, options)`](https://github.com/timgit/pg-boss/blob/master/docs/api/jobs.md), which inserts rows into `pgboss.job`. Workers consume them via [`boss.work(name, options, handler)`](https://github.com/timgit/pg-boss/blob/master/docs/api/workers.md).
+
+> **Note:** pg-boss also has a separate [pub/sub API](https://github.com/timgit/pg-boss/blob/master/docs/api/pubsub.md) (`publish`/`subscribe`) for fan-out to multiple queues. We do **not** use pub/sub — we use `send`/`work` for direct queue processing.
```
BigLep: I think this content better belongs in "Pg-boss vs Dealbot Scheduling" section

docs/jobs.md
```diff
@@ -0,0 +1,216 @@
...
+
+Source: [`jobs.service.ts`](../apps/backend/src/jobs/jobs.service.ts).
+
+## Pg-boss vs Dealbot Scheduling
```
BigLep: I think this section is more important than " Job Types and Queues" for understanding the system.  I would move it up after entities.

BigLep (commented):
Thanks @SgtPooki .  My comments are below.  Maybe it's quicker to verbally discuss/decide on 2026-02-12?  I'm happy to do that.

docs/jobs.md
```diff
@@ -0,0 +1,216 @@
...
+
+- pg-boss provides durable queues and worker execution. Dealbot owns schedule state and timing via `job_schedule_state`.
+- We do not call `boss.schedule(...)` or use pg-boss cron scheduling. The scheduler tick decides when a job is due and enqueues it.
+- This custom scheduler exists because we need per-SP rate schedules, catch-up behavior with caps/jitter, maintenance-window skipping, and a cross-queue per-SP mutex (deal + retrieval).
```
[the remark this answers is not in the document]
SgtPooki: If we want dealbot to give us the real picture of SPs... AFAIK using pg-boss schedules instead of our own logic removes our ability to have per-sp job isolation. I really think spamming SPs with multiple queries is a bad call. Yes, SPs should be able to handle it. But their fluctuation of throughput/timings/etc makes dealbot less trustworthy. By only doing a single job against an SP at a time, we ensure that dealbot isn't:

1. causing more strain on SPs than necessary (impacting other users, other dealbots, etc)
2. showing unnecessary variations in metrics (multiple jobs against an sp could end up "always" showing lower throughput, higher latency, more failures, etc..); If this happens, we have no CLEAR picture WHERE this SP is failing, only that it seems like its "bad" at certain times.

This per sp lock is intended to test each "check" in isolation. If we remove the per sp lock, even if we remove the "catchup", the consequences are less confidence and a devaluation of the assurity in dealbot and the SPs it tests.

docs/jobs.md
```diff
@@ -0,0 +1,216 @@
...
+
+  Workers --> BossJobs
+  Workers --> MutexTable
+  Workers --> DB
+```
+
+## Parallelism and Limits
+
+- **Queue concurrency**: Deal queue concurrency is `DEAL_MAX_CONCURRENCY` (pg-boss [`batchSize`](https://github.com/timgit/pg-boss/blob/master/docs/api/workers.md)), retrieval queue concurrency is `RETRIEVAL_MAX_CONCURRENCY`, and metrics/cleanup are fixed at `batchSize=1`. These limits apply per instance; total concurrency scales with the number of worker pods.
+- **Per-SP mutex**: `job_mutex` ensures only one job runs per SP at a time across both deal and retrieval queues. If a lock is held, the job is re-queued after `JOB_LOCK_RETRY_SECONDS`.
```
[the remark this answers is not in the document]
BigLep: Removed mutex table sounds good.

I don't think we want to use singletonKey.  My understanding is that that is basically an idempotency key for a job.

I think you want to use `groupConcurrency` and set a group per SP.

```
await boss.send('shared-queue', data, {
  group: { id: 'sp1' }
})

await boss.work('shared-queue', {
  groupConcurrency: 1
}, async ([job]) => {
  await processData(job.data)
})
```

If it simplifies implementation, you could have a queue for each job types.  That would mean you could have a retrievalCheck and a dataStorageCheck happening in parallel for an SP, but I don't think that is critical.  Anyways, one queue is fine.

docs/jobs.md
```diff
@@ -0,0 +1,215 @@
...
+
+| Entity | Cardinality | Notes |
+| --- | --- | --- |
+| `job_schedule_state` | One per `<sp, job_type>` plus global rows | Schedule state owned by dealbot. |
+| Storage provider (SP) | One per SP in registry | Filtered by `USE_ONLY_APPROVED_PROVIDERS` when enabled. |
+| Job type | `deal`, `retrieval`, `metrics`, `metrics_cleanup` | `deal` corresponds to "data storage check" externally; we keep `deal` in code/DB for compatibility. |
+| pg-boss queue | `sp.work`, `metrics.run`, `metrics.cleanup` | `sp.work` is a singleton queue. |
+| Dealbot scheduler | One per process (when enabled) | Runs the scheduling loop. |
+| Dealbot worker process | One Node.js process with `DEALBOT_RUN_MODE=worker` or `both` | Hosts pg-boss workers. |
+| pg-boss worker | One per queue per dealbot worker process | Created by each `boss.work(...)` call. |
```
BigLep: I believe the number of pg-boss workers depends on the size of `batchSize`.

docs/jobs.md
```diff
@@ -0,0 +1,215 @@
...
+## Pg-boss vs Dealbot Scheduling
+
+- pg-boss provides durable queues and worker execution.
+- Dealbot owns schedule state and timing via `job_schedule_state` and does not use pg-boss cron scheduling (`boss.schedule(...)`).
+- This custom scheduler exists because we need per-SP rate schedules, controlled backfill with caps/jitter, maintenance-window rules, and per-SP singleton execution across deal + retrieval.
+- `cron` mode is still supported for external users, but pg-boss is the recommended and default operational mode for dealbot-managed deployments.
+
+## Job Types, Queues, and Handlers
+
+| Job type | Queue | Handler | Payload |
```
BigLep: Ideal if the source code links were actual links for easy clicking.

docs/jobs.md
```diff
@@ -0,0 +1,215 @@
...
+
+## How Schedules Are Created and Updated
+
+1. **Startup provider sync**: When pg-boss is enabled, `JobsService` calls `WalletSdkService.ensureWalletAllowances()` and `WalletSdkService.loadProviders()` (unless `DEALBOT_DISABLE_CHAIN=true`). `loadProviders()` pulls providers from the on-chain SP registry and syncs them into `storage_providers`.
+1. **Scheduler tick creates/updates schedules**: The scheduler loop runs immediately on startup and then every `JOB_SCHEDULER_POLL_SECONDS`. It upserts schedules for each active provider and ensures global metrics schedules exist. It deletes deal/retrieval schedules for providers that are no longer active/approved.
+1. **New SP added to registry (example)**: Once `loadProviders()` syncs a new provider, the next scheduler tick inserts `deal` and `retrieval` schedules for that SP.
+1. **SP status changes (example)**: When `loadProviders()` updates provider status, the next scheduler tick re-evaluates and deletes schedules for inactive/unapproved providers.
+
+Note: In pg-boss mode, provider sync currently happens at startup (and whenever `loadProviders()` is called). There is no periodic refresh loop yet.
```
BigLep: Link to the backlog item where this changes?

docs/jobs.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Capacity and Limits
+
+Use these formulas to reason about whether the system can keep up and how much backlog it can absorb.
+
+Per-SP capacity (one job per SP at a time):
+
+- Per-SP execution-minutes per hour = `(deals_per_sp_per_hour * deal_max_minutes) + (retrievals_per_sp_per_hour * retrieval_max_minutes)`
+- If per-SP execution-minutes per hour > 60, that SP can never catch up (backlog grows).
+- If per-SP execution-minutes per hour <= 60, backlog will eventually drain (headroom = `60 - per-SP execution-minutes per hour`).
```
BigLep:
```suggestion
- If per-SP execution-minutes per hour <= 60, backlog should eventually drain, assuming there are enough dealbot workers (headroom = `60 - per-SP execution-minutes per hour`).
```

docs/jobs.md
```diff
@@ -0,0 +1,216 @@
+# Jobs (pg-boss)
+
+This doc explains what a job is in dealbot when `DEALBOT_JOBS_MODE=pgboss`. For operational steps (pausing/resuming/triggering jobs), see the runbook in [`docs/runbooks/jobs.md`](./runbooks/jobs.md).
+
+## Summary
+
+- Jobs are scheduled via `job_schedule_state` rows and executed via pg-boss queues.
+- There is one schedule row per `<job_type, sp_address>` plus global jobs that use an empty `sp_address`.
+- A scheduler loop polls for due schedules, enqueues pg-boss jobs, and advances `next_run_at`.
+- Workers poll pg-boss queues via [`boss.work()`](https://github.com/timgit/pg-boss/blob/master/docs/api/workers.md) and execute the actual deal/retrieval/metrics handlers.
```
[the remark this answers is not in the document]
BigLep: Per verbal it seems like this can be dropped, but it can be done at another point separately.

docs/jobs.md
BigLep: I think we covered a lot of this verbally and anything else still open is covered with other comments so I'll resolve this.

docs/jobs.md
```diff
@@ -0,0 +1,216 @@
...
+
+- `api`: API server + scheduler + `/metrics` (no workers)
+- `worker`: workers + `/metrics` only (no API)
+- `both`: API server + scheduler + workers + `/metrics`
+
+Notes:
+
+- In pg-boss mode, `DEALBOT_RUN_MODE=api` disables workers.
+- Set `DEALBOT_PGBOSS_SCHEDULER_ENABLED=false` to disable the enqueue loop (worker-only pods).
+- `/metrics` is served on `DEALBOT_PORT` for `api`/`both` and on `DEALBOT_METRICS_PORT` for `worker`.
```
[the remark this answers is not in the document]
BigLep: Ok.  I guess we leave for now and it could get simplified later if it's tripping anyone up.  Resolving.

docs/jobs.md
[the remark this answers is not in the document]
BigLep: I agree not all of these requirements can be mapped to `boss.schedule(...)`, especially automatic backfilling.  I'll resolve this comment.

BigLep:
I went through the comments and resolved things I saw got incorporated already or are no longer relevant.  It should just require a pass from you now @SgtPooki to see open comments and respond, incorporate, or resolve.

BigLep:
2026-02-13 conversation notes:
1. Agreed singleton policy is good
2. Russell will add a note about how check statuses relate to pg-boss job statuses
3. Agreed going to do a single concurrency env var (rather than separate for data storage and retrieval).
4. https://github.com/FilOzone/dealbot/blob/biglep/initial-public-explanation-docs/docs/checks/data-storage.md will get udated to have statues `failed.other` and `failed.timedout`.

PR #249, opened 2026-02-10
fix: ipfs data is validated

SgtPooki:
## Summary
Switches IPNI CAR validation from size-only to **ipfs-block** validation. We now verify every block’s multihash by downloading the root block, verifying it, and then walking each child block and verifying the bytes and hashes of each of them.

## Problem
IPNI retrieval previously relied on a byte-count check. A corrupted or tampered CAR could pass validation if its length matched.

### DIFF
--- a/apps/backend/src/retrieval-addons/retrieval-addons.service.ts
+++ b/apps/backend/src/retrieval-addons/retrieval-addons.service.ts
@@ -359,6 +359,9 @@ export class RetrievalAddonsService {
             details: errorMessage,
           };
         }
+        // TODO: totalTime includes per-block hash verification (validateBlock) which
+        // inflates latency and deflates throughput relative to pure network performance.
+        // Splitting fetch and hash into separate queues would give accurate download metrics.
         const totalTime = performance.now() - startTime;
         const responseSize = validation.bytesRead ?? 0;
         const throughput = totalTime > 0 && responseSize > 0 ? responseSize / (totalTime / 1000) : 0;
@@ -368,8 +371,7 @@ export class RetrievalAddonsService {
           method: urlResult.method,
           metrics: {
             latency: Math.round(totalTime),
-            // TODO: decide how to aggregate per-block TTFB for IPFS block fetches.
-            ttfb: 0,
+            ttfb: validation.ttfb ?? 0,
             throughput,
             statusCode: validation.isValid ? 200 : 0,
             timestamp: new Date(),
--- a/apps/backend/package.json
+++ b/apps/backend/package.json
@@ -41,7 +41,6 @@
     "@nestjs/typeorm": "^11.0.0",
     "@willsoto/nestjs-prometheus": "^6.0.2",
     "any-signal": "^4.2.0",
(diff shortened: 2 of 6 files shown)


apps/backend/src/retrieval-addons/retrieval-addons.service.ts
```diff
@@ -344,6 +343,42 @@ export class RetrievalAddonsService {
...
+        const totalTime = performance.now() - startTime;
+        const responseSize = validation.bytesRead ?? 0;
+        const throughput = totalTime > 0 && responseSize > 0 ? responseSize / (totalTime / 1000) : 0;
+
+        return {
+          url: urlResult.url,
+          method: urlResult.method,
+          metrics: {
+            latency: Math.round(totalTime),
+            ttfb: 0,
```
SgtPooki: we need to figure out if we want to emit individual TTFB for each block.. I think yes?

apps/backend/src/retrieval-addons/strategies/ipni.strategy.ts
```diff
@@ -1,23 +1,48 @@
...
+ * Terminology:
+ * - **IPNI** = the index that lists the SP as a provider for content (we do not call an "IPNI URL" here).
+ * - **SP endpoint** = the storage provider's content gateway; we call this to fetch blocks (GET /ipfs/<cid>).
+ *
+ * This strategy validates that the SP serves the root CID and all DAG blocks by traversing links from
+ * the root and fetching each block from the **SP endpoint** with Accept: application/vnd.ipld.raw.
+ * No CAR involved.
  */
 @Injectable()
 export class IpniRetrievalStrategy implements IRetrievalAddon {
```
SgtPooki: this class should be renamed.
```suggestion
export class IpfsBlockRetrievalStrategy implements IRetrievalAddon {
```

once we change this, we can add a "IpfsCarRetrievalStrategy"

apps/backend/src/retrieval-addons/strategies/ipni.strategy.ts
```diff
@@ -62,108 +87,149 @@ export class IpniRetrievalStrategy implements IRetrievalAddon {
...
-   * The PDP provider returns a CAR file, so we validate by checking the size matches
-   * the expected CAR file size from metadata.
-   *
-   * TODO: In the future, we should also:
-   * - Parse the CAR file and extract blocks
-   * - Verify block CIDs match expected blockCIDs
-   * - Reconstruct original data and verify against originalSize
-   * SEE https://github.com/FilOzone/dealbot/issues/144 for more details.
+   * Validate by traversing the DAG from the root CID and fetching each block from the SP endpoint,
+   * not from IPNI. We use GET /ipfs/<cid> with Accept: application/vnd.ipld.raw, verify the CID
```
SgtPooki:
```suggestion
   * Validate by traversing the DAG from the root CID and fetching each block from the SP endpoint.
   * We use GET /ipfs/<cid> with Accept: application/vnd.ipld.raw, verify the CID
```

PR #264, opened 2026-02-11
fix: always disable CDN testing

### LINKED ISSUE #169: remove ENABLE_CDN_TESTING and places that use it
we should remove `ENABLE_CDN_TESTING` dealbot should be testing SPs, not services that wrap SPs

we should also remove all the CDN testing code in the codebase.. I don't think this will come back

SgtPooki:
## Summary
Remove the `ENABLE_CDN_TESTING` environment variable and always disable CDN testing
in deal creation.

## Problem
The `ENABLE_CDN_TESTING` flag controlled whether CDN was randomly enabled (50%
chance) during deal creation. However, dealbot should be testing storage providers
directly — not services that wrap SPs. Having a toggle for CDN testing adds
unnecessary configuration complexity and inconsistent test behavior.

## Solution
- Removed `ENABLE_CDN_TESTING` from the Joi validation schema, config interface
(`IBlockchainConfig`), and config loader
- Hardcoded `enableCDN = false` in `DealService.getTestingDealOptions()`
- Removed the env var from `.env.example`, kustomize configmap, and all
documentation
- Updated test mocks and assertions to reflect the change

## Notes
- The `enableCDN` parameter still flows through `DealPreprocessingConfig` and the
CDN addon strategy — only the config toggle was removed
- Existing deals in the database are unaffected; this only changes future deal
creation behavior
- No migration needed

Fixes #169

### DIFF
--- a/apps/backend/src/database/migrations/1760600000000-RemoveCdnServiceType.ts
+++ b/apps/backend/src/database/migrations/1760600000000-RemoveCdnServiceType.ts
@@ -16,8 +16,8 @@ export class RemoveCdnServiceType1760600000000 implements MigrationInterface {
     await queryRunner.query(`DELETE FROM metrics_daily WHERE service_type = 'cdn'`);
     await queryRunner.query(`
       UPDATE deals
-      SET metadata = metadata - 'cdnMetadata'
-      WHERE metadata ? 'cdnMetadata'
+      SET metadata = metadata - 'cdn'
+      WHERE metadata ? 'cdn'
     `);
     await queryRunner.query(`
       UPDATE deals


apps/backend/src/config/app.config.ts
silent-cipher: `filBeam` config will be of no use after disabling cdn testing completely- Do we still need to keep this around?

SgtPooki:
good catches @silent-cipher , removed with [`cd576e4` (this PR)](https://github.com/FilOzone/dealbot/pull/264/commits/cd576e4ebc242a7afb6b84fb63b11f5d5c0df488) and added a migration to remove from the tables.

PR #266, opened 2026-02-11
fix: remove RETRIEVAL_TIMEOUT_BUFFER_MS

SgtPooki:
built on top of https://github.com/FilOzone/dealbot/pull/263.

we no longer need the RETRIEVAL_TIMEOUT_BUFFER_MS with pgboss workers and retrieval's handling abort signals.


silent-cipher (approved):
LGTM!

PR #268, opened 2026-02-11
fix: providers table is updated every 4 hours

### LINKED ISSUE #260: pgboss/jobs mode should refresh providers daily
i double checked this again, we need a "check for providers" job or something that can query once a day.. walletSdkService currently only loadsProviders on start/restart of a 'worker', 'api', or 'both' node.

_Originally posted by @SgtPooki in https://github.com/FilOzone/dealbot/pull/259#discussion_r2790333942_

SgtPooki:
add daily job to update providers.

previously, providers were only updated on startup or restart.

fixes #260


apps/backend/src/jobs/jobs.service.ts
silent-cipher: In the case of cache-miss, there might be chances of concurrent `loadProviders` calls. Should we consider adding a simple mutex?

silent-cipher (approved):

PR #293, opened 2026-02-18
IPFS validation: ttfb, improved queueing, multiformats for traversal & validation, tests

rvagg:


apps/backend/src/retrieval-addons/strategies/ipfs-block.strategy.ts
```diff
@@ -134,25 +144,23 @@ export class IpfsBlockRetrievalStrategy implements IRetrievalAddon {
     }
     let failed = 0;
     const queue: string[] = [];
-    const enqueued = new Set<string>();
```
rvagg: redundant, we have `seen` so should use it

apps/backend/src/retrieval-addons/strategies/ipfs-block.strategy.ts
```diff
@@ -134,25 +144,23 @@ export class IpfsBlockRetrievalStrategy implements IRetrievalAddon {
...
 
     enqueue(rootCIDStr);
 
     const processCid = async (cidStr: string) => {
-      if (seen.has(cidStr)) return;
-      seen.add(cidStr);
       signal?.throwIfAborted();
 
-      const url = `${spEndpoint}/ipfs/${cidStr}`;
+      const url = `${spEndpoint}/ipfs/${cidStr}?format=raw`;
```
rvagg: not strictly necessary, but spec says SHOULD

Issue #311, opened 2026-02-25
fix: structured logs include providerId / providerAddress / pieceCid / ipfsRootCID where appropriate.


SgtPooki:
the next step after #310

PR #312, opened 2026-02-25
fix: use structured logging

### LINKED ISSUE #310: fix: logs are structured and in one line.
this should make searching/filtering in betterstack much better. and will help with #278 when filtering by SP

SgtPooki:
## Summary
This PR standardizes backend logging to structured JSON-friendly payloads so logs are easier to query in Better Stack and less likely to break into multi-line/unstructured output.

## Changes
- Switched Nest bootstrap logger to JSON output (`ConsoleLogger`, `json: true`, `colors: false`) with explicit `LOG_LEVEL` mapping.
- Added a shared structured error serializer (`toStructuredError`) to normalize thrown values into:
  - `type`, `name`, `message`, `code`, `stack`, `details`
- Replaced legacy error/warn logging patterns across backend services with structured objects, typically:
  - `event`, `message`, `error`
- Removed old string-flattening error formatting usage.
- Updated `filecoin-pin` logger adapter to emit structured entries instead of string-concatenated payloads.
- Updated/added tests for the new logging shape:
  - `apps/backend/src/common/logging.spec.ts`
  - adjusted assertions in existing specs (e.g. utils/wallet-sdk)

## Validation
- `pnpm -C apps/backend typecheck` passed.
- `pnpm -C apps/backend test` passed (`16` files, `158` tests).

## Notes
- This PR focuses on log structure consistency only.
- It does **not** yet add full structured context propagation (`providerId`, `spAddress`, `dealId`, `pieceCid`) to every relevant log line; that remains follow-up work (see #311)
- Expected impact is logging shape only (no intended business-logic behavior change).

Fixes #310


silent-cipher (commented):
One general observation: in a few places we log `errorMessage`, which feels a bit redundant since `toStructuredError(error)` already includes the message in its output.

PR #315, opened 2026-02-27
fix: log datasource startup errors

SgtPooki:
adding on to https://github.com/FilOzone/dealbot/pull/314


apps/backend/src/database/database.module.ts
```diff
@@ -15,6 +18,26 @@ import { StorageProvider } from "./entities/storage-provider.entity.js";
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
 
+// Keep startup diagnostics visible regardless of runtime LOG_LEVEL configuration.
+const startupLogger = new ConsoleLogger("DatabaseModule", {
+  json: true,
+  colors: false,
+  logLevels: [...NEST_STARTUP_LOG_LEVELS],
+});
```
silent-cipher: It might be worth extracting the logger initialization, as it’s currently used in multiple places.

Issue #68, opened 2025-12-02
Set up 7 day database backups


rjan90:
## Description
Implement automated database backup system with 7-day retention or equivalent storage. Current VPS-level backups are insufficient for production reliability.

## Context
- Three key database tables: SP, deals, retrievals
- Currently only VPS-level backups exist
- 7-day retention sufficient to understand SP reliability & trend in case of data corruption.

## Acceptance Criteria
- [x] Automated daily backups configured
- [x] 7-day retention policy implemented
- [x] Monitoring/alerting for backup failures
- [x] Backup restoration process tested
- [x] Backup configuration is documented in team documentation
- [x] Post-GA (M4.5) follow-up issue created for 30-day retention: #348

SgtPooki:
We are planning to use Supabase for Dealbot for now, and all of these items should be easy to complete.

BigLep:
> [@BigLep](https://github.com/BigLep) can you upgrade to pro on our supabase so i can complete this one?

@SgtPooki : done - we are upgraded

SgtPooki:
setup alerts that validate backups with https://github.com/FilOzone/infra/pull/47

BigLep:
2026-02-27 conversation: 
1. we are going to go through the "Backup restoration process tested" to make sure we catch any gotchas
2. We'll add a pointer to https://www.notion.so/filecoindev/FOC-Operational-Excellence-2b7dc41950c1802aa432fff8ecb801cc?source=copy_link#314dc41950c18053b74fe3b0a122432e about how to restore a backup (it's fine if the bulk of this writeup is just linking to Supabase public docs)

PR #337, opened 2026-03-09
docs: add supabase restore runbook

silent-cipher:
## Summary
 
This PR adds a standalone operational script for backing up and restoring Supabase databases, along with a comprehensive runbook for database operations.

### Script Features

- Multiple action modes for different operational needs
  - backup - Create database backup (roles, schema, data)
  - restore - Restore from backup to target database
  - test - Backup + restore validation (creates temporary database)
- Schema scrubbing - automatically replaces supabase_admin with postgres


BigLep:
@silent-cipher : yes, what is happening in this PR is certainly more than I was expecting.  I like the robustness of it, but is more than we outlined in https://github.com/FilOzone/dealbot/issues/68

I was expecting we'd have attempted a restore and have a runbook for how to do it.

My comments earlier were in the context of, "oh ok, if we're going to do something more robust like test the restore process in CI, then how do we find out if it's actally working or not?"  I saw you producing metrics, but how do we make the metrics get collected?

Anyways, I guess I should have started with questioning whether we need this robust of a solution to begin with right now.  (I like the bar that you're trying to hit here, which is why I went along with it, but it does add time to review/maintain, and this is something we can cut.)

Next steps: I think we do one of these:
1. Close this PR and just do the manual items (try a restore and write a runbook) OR
2. Turn this into a script that isn't a nightly CI job that can do a restore.  Then the runbook is basically just: "Run script X at Y."

Does that make sense?

Also, I didn't see @SgtPooki 's comment.  Where was that?

BigLep:
Per before, I think this is good to merge.  Although maybe you wanted to look @SgtPooki ?

PR #339, opened 2026-03-09
fix: data-retention-check counter survives restarts

SgtPooki:
## Problem

While investigating data-retention check counts on BetterStack, I observed ~10,797 `dataSetChallengeStatus` counter increments for a single SP (provider 9) over a 3-hour window. Data-retention checks are based on PDP proofs completed on-chain. On calibnet these happen roughly once per dataset every ~2 hours. Even for an SP with many datasets, 10,797 in 3 hours is wildly inflated.

## Root cause

The `DataRetentionService` tracks cumulative faulted/success proving-period totals per provider in an in-memory `Map` (`providerCumulativeTotals`). On each poll, it queries the PDP subgraph for on-chain proof data, computes the current cumulative totals, and increments Prometheus counters by the delta since the last poll.

The problem: **this map resets to empty on every service restart.** When the first poll runs after a restart, `previous` is `undefined`, so the delta is computed as:

```ts
faultedDelta = estimatedTotalFaulted - 0n  // full cumulative history
successDelta = estimatedTotalSuccess - 0n  // full cumulative history
```

This dumps the entire cumulative proving-period history (potentially hundreds of thousands of periods) into the Prometheus counter as a single massive increment. BetterStack's delta query then correctly picks this up as a huge spike of "new" checks — inflating the reported count every time the service restarts.

This was triggered by a recent deploy that added the `network` label to Prometheus metrics, which caused a pod restart.

## Investigation

I confirmed the issue by querying the PDP subgraph directly for provider 9 (`0xa3971A7234a3379A1813d9867B531e7EeB20ae07`):

- `totalProvingPeriods` (on-chain): 1,014,086
- `totalFaultedPeriods` (on-chain): 93,190
- `estimatedOverduePeriods` (from 100 proof sets): 35,572
- **Total cumulative baseline that would be dumped on restart: ~1,049,658**

The 10,797 observed in BetterStack is a fraction of this full dump, modulated by `avgMerge` time-bucket smoothing.

## Approach

I considered a few options:

1. **Skip counter emission on first poll per provider** (when `previous === undefined`) — simplest, but creates a ~1 hour blind spot after every restart.
2. **Persist baselines to the database** — eliminates the restart inflation entirely without losing data. The table only ever has one row per provider (~21 rows), so overhead is negligible.
3. **Increasing `first:` on the subgraph `proofSets` query** — I also found that the subgraph query silently truncates at 100 proof sets (The Graph's default). However, since the on-chain `totalFaultedPeriods`/`totalProvingPeriods` are already complete aggregate totals, the truncated `estimatedOverduePeriods` self-corrects when proof sets settle on-chain. This is a separate, lower-priority issue.

I went with option 2.

## Solution

Persist the `providerCumulativeTotals` baselines to a new `data_retention_baselines` database table so they survive service restarts. On startup, baselines are loaded from the DB before any polling occurs. If the DB is unavailable, the poll is skipped entirely (rather than emitting bloated values) and retried on the next tick.

## Impact

- **Restarts no longer inflate metrics.** On startup, baselines are loaded from the DB, so the first poll computes a real delta (typically near-zero if the restart was quick) instead of re-emitting the full history.
- **No additional subgraph API calls.** The fix is entirely on the persistence side.
- **Safe degradation.** If the DB is unavailable at startup, the poll is skipped and retried on the next tick. No data is emitted until baselines are successfully loaded.
- **Minimal DB overhead.** The table stores one row per active provider (~21 rows). Upserts happen once per provider per poll (default: hourly).

### DIFF
--- a/docs/checks/data-retention.md
+++ b/docs/checks/data-retention.md
@@ -81,6 +81,8 @@ faultedDelta = currentTotalFaulted - previousTotalFaulted
 successDelta = currentTotalSuccess - previousTotalSuccess
 ```
 
+**First-seen provider handling**: When a provider has no prior baseline (fresh deploy or newly added provider), dealbot initializes the baseline to the current cumulative totals **without emitting any counters**. This prevents dumping the provider's full cumulative history as a single metric spike. Metrics for that provider will begin accumulating from the next poll onward.
+
 **Negative delta handling**: If deltas are negative (due to chain reorgs, subgraph corrections, or data inconsistencies), the baseline is reset to current values without incrementing counters. This prevents stalled metrics.
 
 **Baseline persistence**: Baselines are persisted to the `data_retention_baselines` database table after each successful poll. On service restart, baselines are reloaded from the database to prevent metric inflation.
@@ -104,8 +106,9 @@ To prevent metric inflation across service restarts, dealbot persists provider b
 **Lifecycle**:
 
 1. **On first poll**: Load all baselines from database into memory. If load fails, abort poll to prevent emitting inflated values.
-2. **On each poll**: After processing providers, persist updated baselines to database.
-3. **On restart**: Reload baselines from database. Delta computation resumes from last persisted state, preventing double-counting.
+2. **First-seen provider**: If a provider has no prior baseline (not in memory or database), initialize its baseline to the current cumulative totals without emitting counters. This avoids a metric spike from the provider's full history.
+3. **On each poll**: After processing providers, persist updated baselines to database.
+4. **On restart**: Reload baselines from database. Delta computation resumes from last persisted state, preventing double-counting.
 
 **Error handling**:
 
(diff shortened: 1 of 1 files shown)


silent-cipher:
> I think this is too strict. If at time=0 after a restart we are unable to load from the DB the previous value but we were able to get a value from the subgraph, that should be our new baseline, then at time=0+checkPeriod we can look at the delta from the value we got from the subgraph at time=0 vs. the value we got at time=0+checkPeriod.

I think this approach will miss some counters.
Assume at time=0 after a restart we updated the baseline with value from the subgraph without increasing the counters. In that case, the delta between the previous baseline and the subgraph value will be lost forever, since we updated the baseline without accounting for those counters.

apps/backend/src/data-retention/data-retention.service.ts
```diff
@@ -279,8 +320,29 @@ export class DataRetentionService {
...
+        event: "baseline_initialized",
+        message: `Initialized baseline for provider ${address} (no prior baseline)`,
+        providerAddress: address,
+        providerId: providerInfo.id,
+        faultedPeriods: estimatedTotalFaulted.toString(),
+        successPeriods: estimatedTotalSuccess.toString(),
+      });
+      this.providerCumulativeTotals.set(normalizedAddress, newBaseline);
+      return newBaseline;
+    }
```
silent-cipher: Won’t this cause us to miss some counter increments?
Assume we skip a provider with 5 faulted and 100 success periods, and instead of emitting counters we just set this as the baseline. In the next loop, if we observe 7 faulted and 120 success periods, the emitted counters would only be 2 faulted and 20 success periods. In this case, the initial baseline counters (5 faulted, 100 success periods) are never emitted and effectively lost. Is that correct?

PR #344, opened 2026-03-10
fix: ipni verification polls for at least 60 seconds

SgtPooki:
## Summary
This PR improves IPNI verification reliability and observability for data-storage checks that were failing with `root CID not verified`.

## Investigation Context
Storage providers reported repeated IPNI verification failures.
Reference thread: [Slack investigation thread](https://filecoinproject.slack.com/archives/C08TVNKJV7C/p1773081058305829)

In BetterStack, we observed failures like:
- `IPNI verification failed for deal ...: root CID not verified`
- `ipni_root_cid_verification_failed` after a short verification window

Reference query: [BetterStack tail link](https://telemetry.betterstack.com/team/t468215/tail?s=1678393&q=message.message.providerId%3D%229%22%20AND%20bafybeiaiw4khpjov3jtor3crso5ig2g6x23vw7opzqgwaz5ivloddjzyhe&a=1773151863954703.89499195)

## Root Cause
IPNI verification runtime could be shorter than expected due to retry-budget math.
Previous logic divided the attempt count by 2 (`ATTEMPT_MULTIPLIER`), effectively halving the retry window and causing early verification failure under normal propagation delay.

### 1. Timeout + Retry Handling
- Removed the `ATTEMPT_MULTIPLIER` that halved retry attempts.
- New retry budget: `maxAttempts = ceil(timeoutMs / delayMs) + 1`.
- Verification is now bounded by both:
  - An internal wall-clock timeout (`AbortSignal.timeout(IPNI_VERIFICATION_TIMEOUT_MS)`)
  - The parent deal/job signal (`AbortSignal.any([...])`)

### 2. Timeout Default
- Increased `IPNI_VERIFICATION_TIMEOUT_MS` default from `10000` to `60000` in config and docs.

### 3. Structured Logging
Added IPNI verification log events at appropriate levels:
- `ipni_verification_started` (log)
- `ipni_verification_timed_out` (error)
- `ipni_verification_failed` (error)
- `ipni_verification_succeeded` (log)

These include context fields: `rootCID`, `providerAddress`, `providerId`, `serviceUrl`, `expectedMultiaddr`, `blockCIDCount`, `timeoutMs`, `pollIntervalMs`, `maxAttempts`, and `failureReason` (on failures).

Also enriched strategy-level `ipni_root_cid_verification_failed` logs with `failureReason` and `failedCIDs`.

### 5. Verification Start Point (unchanged)
IPNI verification still starts after SP piece status reaches `advertised`.

## Why This Approach
- Makes runtime behavior match configured timeout and polling expectations.
- Preserves cancellation guarantees from the parent job signal.
- Adds sufficient log context to distinguish propagation delay, provider mismatch, and timeout scenarios during incident triage.

## Expected Impact
- Fewer false-negative IPNI failures from too-short retry windows.
- Faster incident diagnosis from richer structured error logs.
- Longer but bounded verification window for slow IPNI propagation, still constrained by job-level abort.

## Options Considered
- **Only increasing timeout** without fixing retry math — rejected; retry behavior remained hard to reason about.
- **Using filecoin-pin defaults with only a signal** — rejected; less explicit control and harder to predict behavior.
- **Delaying verification until `piece_confirmed`** — not in scope; current flow intentionally remains post-`advertised`.

## Testing

All 20 test suites passed (271 tests).

## Notes
- No DB migrations.
- No API contract changes.
- New structured IPNI log events can be used directly in BetterStack dashboards and alert queries.

### DIFF
--- a/apps/backend/src/ipni/ipni-verification.service.ts
+++ b/apps/backend/src/ipni/ipni-verification.service.ts
@@ -10,7 +10,7 @@ export type IpniVerificationInput = {
   blockCids?: CID[];
   storageProvider: StorageProvider;
   timeoutMs: number;
-  pollIntervalMs?: number;
+  pollIntervalMs: number;
   signal?: AbortSignal;
 };
 
@@ -26,10 +26,11 @@ export class IpniVerificationService {
     pollIntervalMs,
     signal,
   }: IpniVerificationInput): Promise<IPNIVerificationResult> {
-    const delayMs = Math.max(100, pollIntervalMs ?? 2000);
+    const delayMs = Math.max(100, pollIntervalMs);
     // Keep retrying at the configured polling cadence until timeout or outer job cancellation.
-    // Compute maxAttempts from timeoutMs/delayMs and add 1 to avoid rounding down and ensure
-    // the polling loop can span the full timeout window with one extra attempt.
+    // waitForIpniProviderResults uses attempt-then-delay: the first attempt is immediate,
+    // so N attempts span only (N-1) delays. Adding 1 ensures the attempt budget covers
+    // the full timeout window rather than falling short by up to one delayMs interval.
     const maxAttempts = Math.max(1, Math.ceil(timeoutMs / delayMs) + 1);
     const expectedProviders = [this.buildExpectedProviderInfo(storageProvider)];
     const expectedMultiaddr = serviceURLToMultiaddr(storageProvider.serviceUrl);
--- a/apps/backend/src/deal-addons/strategies/ipni.strategy.ts
+++ b/apps/backend/src/deal-addons/strategies/ipni.strategy.ts
@@ -45,7 +45,6 @@ export class IpniAddonStrategy implements IDealAddon<IpniMetadata> {
   readonly priority = AddonPriority.HIGH; // Run first to transform data
   readonly POLLING_INTERVAL_MS = 2500;
   readonly POLLING_TIMEOUT_MS = 2 * 60 * 1000; // 2 minutes - max time to wait for SP to advertise piece
-  readonly IPNI_LOOKUP_TIMEOUT_MS = 3 * 60 * 1000; // 3 minutes - max time to wait for IPNI propagation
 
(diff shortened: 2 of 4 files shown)


apps/backend/src/config/app.config.ts
BigLep: Not introduced with this change but it seems weird to define the default in two places.  Maybe a refactor for another time, but it would be nice to set it once - not two places.  Maybe ask AI to handle that?

apps/backend/src/ipni/ipni-verification.service.ts
```diff
@@ -29,27 +26,93 @@ export class IpniVerificationService {
     pollIntervalMs,
     signal,
   }: IpniVerificationInput): Promise<IPNIVerificationResult> {
-    const intervalMs = Math.max(100, pollIntervalMs ?? DEFAULT_POLL_INTERVAL_MS);
-    const maxAttempts = Math.max(1, Math.ceil(timeoutMs / intervalMs / ATTEMPT_MULTIPLIER));
+    const delayMs = Math.max(100, pollIntervalMs ?? 2000);
+    // Keep retrying at the configured polling cadence until timeout or outer job cancellation.
+    // Compute maxAttempts from timeoutMs/delayMs and add 1 to avoid rounding down and ensure
+    // the polling loop can span the full timeout window with one extra attempt.
+    const maxAttempts = Math.max(1, Math.ceil(timeoutMs / delayMs) + 1);
```
BigLep: If we're taking the ceiling, why do we need to add one.  Doesn't ceiling ensure we aren't rounding down?  Is this this "+1 behavior" standard in retry logic?  I don't remember seeing it.  If not, I'd remove it.

SgtPooki: it could be possible that polling timeout and delay are set such that the max attempts does not 'use up' the max timeout fully.. my thinking was that we want to always have max attempts set to an amount that would always result in a number of attempts that would allow the signal to throw

PR #346, opened 2026-03-10
feat: add per-SP piece cleanup job to bound storage growth

### LINKED ISSUE #281: Cleanup old pieces
## Done Criteria
Dealbot ensures that the amount of stored data stays below a configurable threshold.  When dealbot-owned datasets store more data than this threshold, there is a process that cleans up and produces headroom so more pieces can be added while still keeping the total data stored by the SP below the threshold.  

## Why Important
Currently the dealbot just keeps adding pieces.  After a certain point, this extra data storage load isn't providing any value for what dealbot is trying to check, and it increases cost for the SP. 

## User/Customer
SP

## Notes
1. Design decision need to be made on whether cleanup is part of a "data storage check" or a separate process.
2. When this impleted, `production-configuration-and-approval-methodology.md` should get updated.

Chaitu-Tatipamula:
## Summary

Adds a piece_cleanup job type that periodically deletes the oldest pieces from each SP once total stored data exceeds a configurable quota (`MAX_DATASET_STORAGE_SIZE_BYTES`, default 24 `GiB`). Cleanup runs on the existing sp.work queue with singleton-per-SP semantics, ensuring it doesn't overlap with deal/retrieval work.

Closes #281

## What it does

- **Quota-based cleanup**: Deletes oldest pieces (`FIFO`) until SP is back under quota
- **Over-quota gating**: Skips new deal creation for SPs that exceed the storage quota
- **Idempotent deletion**: Treats contract reverts for already-deleted pieces as success and continues
- **Runtime cap**: Aborts cleanup after `MAX_PIECE_CLEANUP_RUNTIME_SECONDS` (default 300s)
- **Rate-based scheduling**: `JOB_PIECE_CLEANUP_PER_SP_PER_HOUR` (default 1), consistent with `DEALS_PER_SP_PER_HOUR` pattern

### Config & Entity
- New env vars: `MAX_DATASET_STORAGE_SIZE_BYTES`, `JOB_PIECE_CLEANUP_PER_SP_PER_HOUR`, `MAX_PIECE_CLEANUP_RUNTIME_SECONDS`
- Deal entity: `cleaned_up` (bool) and `cleaned_up_at` (timestamptz) columns
- piece_cleanup added to `JobType` union

### Core Service (`piece-cleanup/`)
- `PieceCleanupService` — cleanup loop, quota check, Synapse SDK integration
- `isProviderOverQuota()` — used by deal handler to gate new deals
- Shared Synapse instance with on-demand fallback
- No-progress bail-out to prevent infinite loops when all deletions fail

### Jobs Integration (`jobs/`)
- `handlePieceCleanupJob` with AbortController timeout
- Over-quota check in `handleDealJob`
- Rate-based interval in `getIntervalSecondsForRates`
- Always-scheduled in `ensureScheduleRows`

### Documentation
- `docs/environment-variables.md` — 3 new env vars
- `docs/jobs.md` — piece_cleanup in job types table, capacity formula
- `docs/runbooks/jobs.md` — pause/resume/trigger SQL examples
- `docs/checks/production-configuration-and-approval-methodology.md` — updated FAQ

## Testing

- 24 new tests in `piece-cleanup.service.spec.ts`
- Updated `jobs.service.spec.ts` for new config structure
- **292/292 tests pass**, biome ci clean

## things to be considered

1. **Soft-delete vs hard-delete**: Currently soft-deletes deal rows (cleaned_up=true).
2. **Failure backoff**: Currently logs piece_cleanup_no_progress and retries next hour. maybe we can alert when this event appears like 3+ times in a cleanup interval (1 hour)


BigLep:
NOte we may be delayed looking at this since it's not a priority for the M4.1 mainnet readiness milestone we're working on.

SgtPooki (changes requested):
@Chaitu-Tatipamula the main concern I have with this approach is that we’re using our DB as the source of truth for how much data an SP currently has stored. I think that needs to be inverted. But there are two other main concerns below as well

We should use `synapse-sdk` / `filecoin-pin` first to determine the provider’s actual live stored size, and only if the provider is over quota should we query our DB to decide which pieces are eligible for cleanup. The DB is useful for candidate selection and bookkeeping, but it shouldn’t be the thing that decides whether we’re over quota, because it can drift from chain/provider reality.

I’d also want the cleanup path to reconcile DB state after deletion so we don’t keep records around for pieces that are no longer live. That matters not just for quota accounting, but also because those rows are what we use to decide what retrieval work to run. If a piece is already gone from the provider, we should clean up the DB state as part of reconciliation rather than continuing to treat it as stored data.

Concretely, I think the flow should be:
1. Ask Synapse/filecoin-pin for the SP’s current live usage.
2. If over quota, query the DB for the oldest eligible pieces to delete.
3. After each successful delete, update/remove the DB record so retrieval scheduling stays correct.
4. Add tests around DB/provider drift so we don’t block deals or schedule retrievals based on stale rows.

---

I don’t think this should run hourly by default. #281 doesn’t require that cadence, and I’d prefer cleanup to run once per day unless we have evidence we need it to be more aggressive. The current PR models this as runs-per-hour with a default of 1, which feels too frequent for a maintenance/reconciliation job.

---

I think this needs a second threshold, not just a max quota. When we detect an SP is over quota, we should clean down to a configurable target or create a configurable amount of free headroom, not stop at “barely under quota.” That target needs to be chosen relative to the cleanup cadence and expected growth rate. Otherwise a daily/weekly cleanup can end up either thrashing near the threshold or spending an unbounded amount of time deleting pieces in one run. I’d like to see configurable high-water/low-water behavior here, plus a bounded per-run cleanup budget.

apps/backend/src/piece-cleanup/piece-cleanup.service.ts
```diff
@@ -0,0 +1,376 @@
...
+
+    const datasets = await listDataSets(synapse, {
+      filter: (ds) => ds.serviceProvider === spAddress,
+    });
+
+    if (datasets.length === 0) {
+      this.logger.debug({
+        event: "piece_cleanup_no_live_datasets",
+        message: `SP ${spAddress}: no live datasets found on provider`,
+        spAddress,
```
silent-cipher:
```suggestion
        message: "SP has no live datasets",
        providerAddress: spAddress,
```

apps/backend/src/piece-cleanup/piece-cleanup.service.ts
```diff
@@ -0,0 +1,376 @@
...
+      });
+      return 0;
+    }
+
+    const result = await calculateActualStorage(synapse, datasets, { signal });
+
+    this.logger.debug({
+      event: "piece_cleanup_live_storage_queried",
+      message: `SP ${spAddress}: live storage = ${this.formatBytes(Number(result.totalBytes))} across ${result.dataSetCount} datasets (${result.pieceCount} pieces)`,
+      spAddress,
```
silent-cipher:
```suggestion
      providerAddress: spAddress,
```

silent-cipher (changes requested):
We generally use standardized field names in structured logs to help with filtering in SPs:

- `providerAddress`
- `providerId`
- `providerName`

It would be good to align logs with these conventions. I’ve added a few suggestions, but it may be worth applying this consistently across all log statements.

---

The current deletion flow looks potentially incorrect.
When creating a storage context, the sdk resolves the dataset using either:

- dataset metadata (via options), or
- `dataSetId` (preferred for explicit resolution)

In this case, since no metadata is provided, the SDK may resolve a dataset with empty metadata, which is risky. We should explicitly pass `dataSetId` to ensure correct dataset resolution before calling `deletePiece`.

apps/backend/src/piece-cleanup/piece-cleanup.service.ts
```diff
@@ -0,0 +1,376 @@
...
+    if (datasets.length === 0) {
+      this.logger.debug({
+        event: "piece_cleanup_no_live_datasets",
+        message: `SP ${spAddress}: no live datasets found on provider`,
+        spAddress,
+      });
+      return 0;
+    }
+
+    const result = await calculateActualStorage(synapse, datasets, { signal });
```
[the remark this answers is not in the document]
Chaitu-Tatipamula: hmm we do scope it that way but, `calculateActualStorage` itself just sums the pieces across whatever datasets array you pass it. the client address scoping actually happens at `listDataSets` a step before this, which fetches datasets owned by the synapse client's wallet and then filters by SP. but the end result is as you describe.

silent-cipher (changes requested):

silent-cipher (changes requested):
`job-schedule.repository.ts` likely needs to be revisited.

- `deleteSchedulesForInactiveProviders` should include the `piece_cleanup` job type
- `countBossJobStates` should also account for `piece_cleanup` jobs
- `minBossJobAgeSecondsByState` should include piece_cleanup in its query as well

Chaitu-Tatipamula:
> `job-schedule.repository.ts` likely needs to be revisited.
> 
> * `deleteSchedulesForInactiveProviders` should include the `piece_cleanup` job type
> * `countBossJobStates` should also account for `piece_cleanup` jobs
> * `minBossJobAgeSecondsByState` should include piece_cleanup in its query as well

i am adding `piece_cleanup` to `deleteSchedulesForInactiveProviders` since this hardcodes the jobType. for `countBossJobStates` and `minBossJobAgeSecondsByState`, these already handle it: `piece_cleanup` jobs go through the shared sp.work queue with jobType being "piece_cleanup" wdyt @silent-cipher ?

silent-cipher (approved):
Overall, this lgtm!

One last suggestion: could we pass a consistent log context throughout the piece cleanup process, similar to what we do for deal, retrieval, and dataset creation flows?
This would help keep logging consistent and make debugging easier.

apps/backend/src/piece-cleanup/piece-cleanup.service.ts
```diff
@@ -4,7 +4,7 @@ import { ConfigService } from "@nestjs/config";
 import { InjectRepository } from "@nestjs/typeorm";
 import { calculateActualStorage, listDataSets } from "filecoin-pin/core/data-set";
 import { IsNull, Not, Repository } from "typeorm";
-import { toStructuredError } from "../common/logging.js";
+import { type PieceCleanupLogContext, toStructuredError } from "../common/logging.js";
```
silent-cipher:
```suggestion
import { type PieceCleanupLogContext, type ProviderJobContext, toStructuredError } from "../common/logging.js";
```

apps/backend/src/piece-cleanup/piece-cleanup.service.ts
```diff
@@ -118,30 +120,38 @@ export class PieceCleanupService implements OnModuleInit, OnModuleDestroy {
    * 5. Mark the deal record as cleaned up
    * 6. Repeat until usage drops below TARGET or runtime cap is reached
    */
-  async cleanupPiecesForProvider(spAddress: string, signal?: AbortSignal): Promise<CleanupResult> {
+  async cleanupPiecesForProvider(
+    spAddress: string,
+    signal?: AbortSignal,
+    logContext?: PieceCleanupLogContext,
```
silent-cipher:
```suggestion
    logContext?: ProviderJobContext,
```

silent-cipher (approved):

silent-cipher:
@SgtPooki , when you get a chance, could you take a look at this? There are a few requested changes.

SgtPooki:
> Over-quota gating: Skips new deal creation for SPs that exceed the storage quota


we should *not* prevent deal creation.. we just need to make sure cleanup is releasing the pressure of large storage. 

dealbot should constantly be running jobs that perform data-storage and retrieval checks.. im fixing the merge conflicts and addressing some things and will push up on this PR shortly.

SgtPooki (approved):

Issue #345, opened 2026-03-10
fixup: dealbot SP logs dashboard on betterstack


SgtPooki:
things to fix:

1. some logs are showing unstructured data.
2. we should display columns for each field in the json instead of one json blob for "log_message"

PR #352, opened 2026-03-12
fix: more structured logs

SgtPooki:
addressing high priority items from https://github.com/FilOzone/dealbot/issues/345#issuecomment-4049476099

### DIFF
--- a/apps/backend/src/retrieval-addons/retrieval-addons.service.ts
+++ b/apps/backend/src/retrieval-addons/retrieval-addons.service.ts
@@ -152,6 +152,8 @@ export class RetrievalAddonsService {
           message: `Failed to construct URL with strategy ${strategy.name}`,
           strategy: strategy.name,
           dealId: config.deal.id,
+          providerId: config.deal.storageProvider?.providerId,
+          providerAddress: config.storageProvider,
           error: toStructuredError(error),
         });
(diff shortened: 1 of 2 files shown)


apps/backend/src/retrieval-addons/strategies/ipfs-block.strategy.ts
```diff
@@ -216,7 +243,13 @@ export class IpfsBlockRetrievalStrategy implements IRetrievalAddon {
           .catch((err) => {
             const msg = err instanceof Error ? err.message : String(err);
             failed += 1;
-            this.logger.warn(`IPFS block-fetch validation error for deal ${config.deal.id}: cid ${cidStr} - ${msg}`);
+            this.logger.warn({
+              ...logContext,
+              event: "block_fetch_validation_error",
+              message: "Block fetch failed for CID",
+              cid: cidStr,
+              error: msg,
```
silent-cipher: Should we use `toStructuredError(err)` instead of converting error to a simple string message?

apps/backend/src/retrieval-addons/retrieval-addons.service.ts
silent-cipher: Can we add provider id and address to this log as well?

PR #360, opened 2026-03-13
fix: create data set job only creates 1 per run

SgtPooki:
many data_set_creation_job_aborted errors were showing for providers, so i'm reducing the chance for a timeout to happen by only creating one new dataset per job run per SP.. eventually, all data-sets needed will be created and we won't conflate a failed data-set create job with only one of many failing.

### DIFF
--- a/apps/backend/src/jobs/data-set-creation.handler.ts
+++ b/apps/backend/src/jobs/data-set-creation.handler.ts
@@ -61,21 +61,6 @@ export async function provisionNextMissingDataSet(
     await dealService.createDataSetWithPiece(spAddress, metadata, signal);
     logger.log({
       ...logContext,
-      event: "created_provisioned_data_set",
-      message: "Created provisioned data-set",
-    });
-
-    logger.log({
-      ...dataSetLogContext,
-      event: "data_set_provisioning_progress",
-      message: "Created 1 data-set, deferring remaining to next run",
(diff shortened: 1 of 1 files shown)


apps/backend/src/jobs/data-set-creation.handler.ts
```diff
@@ -60,15 +64,26 @@ export async function provisionDataSets(
       event: "created_provisioned_data_set",
       message: "Created provisioned data-set",
     });
```
silent-cipher: This log seems redundant here.

apps/backend/src/jobs/data-set-creation.handler.ts
```diff
@@ -60,15 +64,26 @@ export async function provisionDataSets(
...
+      checkedExistingCount: existingCount,
+      uncheckedCount: minDataSets - existingCount - 1,
+    });
+    return;
   }
 
   logger.log({
     ...dataSetLogContext,
     event: "data_sets_provisioning_completed",
     message: "Data-set provisioning completed for provider",
```
silent-cipher:
```suggestion
    message: "All required data-sets exist for provider",
```

silent-cipher (approved):
Looks good to me! Nothing blocking, just a few suggestions to improve logging.

PR #362, opened 2026-03-16
fix: log when ipniVerifyMs is > 5s

SgtPooki:
adds a log if ipniVerifyMs time is greater than 5 seconds, because it's usually a sign of a problem, and we would need to correlate logs to find the root cause.

slack chat: https://filecoinproject.slack.com/archives/C07CGTXHHT4/p1773683302799539


BigLep (commented):
@SgtPooki : I want to make sure I'm following what we hope to acomplish with this.  

Basically, are we worried that there is a time gap between `spAnnouncedAdvertisementToIpni` and `ipniVerificationStart` which is making `ipniVerifyMs` not a good metric for "how long it takes for a CID to be verified on filecoinpin.contact?

In practice, I would hope that the timestamp of `spAnnouncedAdvertisementToIpni` and `ipniVerificationStart` are effectively equivalent (i.e., that when `spAnnouncedAdvertisementToIpni` fires that `ipniVerificationStart` also does).

PR #370, opened 2026-03-17
fix: add typeorm bigint column

SgtPooki:
See https://github.com/typeorm/typeorm/issues/11981#issuecomment-4077353177 for context

@juliangruber I needed these changes to get your PR #186 working with `make up`


juliangruber (approved):
Thank you! I had no idea this was the typeorm behavior for bigints, it's surprising indeed.

We should eventually fix that an issue like this isn't caught by the test suite, what do you think about running library unit tests against the database?

PR #375, opened 2026-03-18
test: validate typeorm entities

SgtPooki:
## Summary

- Adds a `postbuild` script that validates compiled TypeORM entity metadata against the Postgres driver, catching unsupported column types before deployment
- Runs automatically after every `pnpm build` — locally, in CI (`test.yml`), and in Docker image builds (`Dockerfile`)

## Problem

PR #186 introduced a `bigint` TS type on `StorageProvider.providerId` without an explicit `type` in `@Column()`. TypeORM can't infer a Postgres column type for JS `BigInt`, so the app throws `DataTypeNotSupportedError` at startup and crash-loops.

## Why unit tests didn't catch it

Vitest uses SWC to transpile, which does not emit `emitDecoratorMetadata` the same way `tsc` does. TypeORM never sees the `BigInt` design type in tests — only in the production `tsc` build output.

## Approaches tried

1. **DataSource.initialize() in vitest** — SWC doesn't emit decorator metadata, so TypeORM validation passes in tests but fails in prod.
2. **Static analysis of entity source files** — regex-based scan for `bigint` types missing explicit `@Column({ type })`. Works but brittle and doesn't cover other validation TypeORM performs.
3. **Postbuild validation of compiled `dist/` output** (chosen) — imports tsc-compiled entities, uses TypeORM's `DataSource.buildMetadatas()` to run the same driver-level validation as production startup, without needing a database connection.

## Test plan

- [x] `pnpm -C apps/backend build` fails on PR #186 with: `✗ entity-metadata: Data type "BigInt" in "StorageProvider.providerId" is not supported by "postgres" database.`
- [x] After fixing the entity (e.g. `@Column({ type: "bigint" })`), build passes with: `✓ entity-metadata`
- [x] `pnpm test` unaffected (275 tests pass)

## Related

- https://github.com/FilOzone/dealbot/pull/370

### DIFF
--- a/apps/backend/scripts/postbuild-checks.mjs
+++ b/apps/backend/scripts/postbuild-checks.mjs
@@ -82,7 +82,7 @@ async function checkEntityMetadata() {
   // requiring a live database connection.
   await ds.buildMetadatas();
 }
-check.push({ name: "entity-metadata", fn: checkEntityMetadata })
+checks.push({ name: "entity-metadata", fn: checkEntityMetadata });
 
 // ---------------------------------------------------------------------------
 // Runner


apps/backend/scripts/postbuild-checks.mjs
```diff
@@ -0,0 +1,107 @@
...
+  // requiring a live database connection.
+  await ds.buildMetadatas();
+}
+
+// ---------------------------------------------------------------------------
+// Runner
+// ---------------------------------------------------------------------------
+
+/** @type {Array<{ name: string; fn: () => Promise<void> }>} */
+const checks = [{ name: "entity-metadata", fn: checkEntityMetadata }];
```
juliangruber:
```suggestion
/** @type {Array<{ name: string; fn: () => Promise<void> }>} */
const checks = [];

/**
 * Validates that every entity's column types are supported by the target
 * database driver.  This catches e.g. using a bare `bigint` TS type on a
 * @Column() without an explicit `type`, which makes TypeORM emit a
 * DataTypeNotSupportedError at startup.
 */
async function checkEntityMetadata() {
  const { DataSource } = await import("typeorm");
  const entities = await loadEntities();

  const ds = new DataSource({
    type: "postgres",
    host: "localhost",
    database: "unused",
    entities,
    synchronize: false,
  });

  // buildMetadatas() validates column types against the driver without
  // requiring a live database connection.
  await ds.buildMetadatas();
}
check.push({ name: "entity-metadata", fn: checkEntityMetadata })

// ---------------------------------------------------------------------------
// Runner
// ---------------------------------------------------------------------------
```

separates checks definition from runner logic

PR #380, opened 2026-03-19
fix: make nullable entity fields explicitly typed

### LINKED ISSUE #205: Fix nullable entity types
Just flagging - we should fix these types for `nullable` fields by marking them as optional

_Originally posted by @silent-cipher in https://github.com/FilOzone/dealbot/pull/175#discussion_r2758951777_

---

we need to do this across all entity types

silent-cipher:
closes #205


SgtPooki (changes requested):
there are a lot of merge conflicts, and this PR has a lot of changes that i haven't got to look into deeply..

I would much rather see a smaller change since we're messing with metrics we're exporting

SgtPooki (approved):

PR #409, opened 2026-03-31
feat: support custom RPC endpoint via RPC_URL

rvagg:
Allows configuring a dedicated RPC endpoint (e.g. authenticated Glif/Chain.Love) to avoid public endpoint rate limiting that causes dataset resolution failures and balance drain.

Doesn't require an RPC_URL to run, but will use one if supplied.

I've not added a separate API key env var because chain.love (etc.) allow it in the URL with `?token=...` ([see](https://chain-love.gitbook.io/chain.loves-api-management-module/getting-started-with-rpc/send-requests)). Unless someone has a good reason to split them apart and set up a custom transport with a bearer token?


BigLep (approved):
Change looks good and makes sense to me. I assume you'll review/resolve copilot coments.

Issue #275, opened 2026-02-13
Remove Metrics table and job


BigLep:
TODO: fill in the details here, but this is about removing the metrics tabls and job since we'll rely on Pometheus to scrape these metrics and another tool to visualize it (see https://github.com/FilOzone/dealbot/issues/239 ).

BigLep:
@SgtPooki : I agree this is good to do, but I'm going to bump to M4.2 since I don't think it's as high of a priority.  (That said, we should do it sooner if it is slowing down other M4.1 work.)

SgtPooki:
> the wallet_balance, storage_providers_active and storage_providers_tested gauges 
> lightweight @nestjs/schedule

Decoupling from pg-boss is good; @nestjs/schedule is unnecessary here. For these specific metrics, a scheduler/interval is still not the best fit. They are point-in-time gauges defined in metrics-prometheus.module.ts (line 203), and prom-client explicitly recommends scrape-time collect() for that kind of metric instead of updating them from setInterval timers, [prom-client docs](https://github.com/siimon/prom-client/blob/2160804545f371906387f91143060a935de0a136/README.md?plain=1#L114-L120)). An interval-based Nest scheduler would make the values inherently stale between ticks, do work even when nobody is scraping, and add another background lifecycle to reason about per process/replica.

We should use a `collect` method and keep a small TTL for wallet_balance so we're not hitting the chain on every scrape. providers_active and providers_tested can be updated as we run the `refresh providers` existing job

silent-cipher:
Completed by #436

PR #436, opened 2026-04-07
refactor: remove metrics job scheduler

silent-cipher:
## Summary

This PR removes the metrics job scheduling system from the backend, eliminating the periodic metrics aggregation and cleanup jobs that were previously managed by pg-boss. The metrics calculation functionality remains available but is no longer automatically scheduled.

## Breaking Changes
- Environment Variable: `METRICS_PER_HOUR` is no longer recognized (can be safely removed from `.env` files)
- Job Types: Existing `metrics` and `metrics_cleanup` jobs in pg-boss will no longer be processed

Implements steps 4, 5 and 10 from https://github.com/FilOzone/dealbot/issues/275#issuecomment-4069535483


apps/backend/src/database/entities/job-schedule-state.entity.ts
```diff
@@ -1,13 +1,16 @@
...
   | "deal"
   | "retrieval"
   | "data_set_creation"
-  | "metrics"
-  | "metrics_cleanup"
   | "providers_refresh"
-  | "data_retention_poll";
+  | "data_retention_poll"
+  | "metrics"
+  | "metrics_cleanup";
```
SgtPooki: unnecessary change here. the comment is useful, but we should add a suffix comment to the end of the ones that should no longer be ran..

and shouldn't a migration take care of this or do we get type errors if we remove these in the same commit as the migration?

we should have a follow up issue for removing them fully

apps/backend/src/jobs/repositories/job-schedule.repository.ts
```diff
@@ -211,12 +209,10 @@ export class JobScheduleRepository {
       SELECT
         CASE
           WHEN name = $2 THEN COALESCE(data->>'jobType', 'unknown')
-          WHEN name = $3 THEN 'metrics'
-          WHEN name = $4 THEN 'metrics_cleanup'
-          WHEN name = $5 THEN 'deal'
-          WHEN name = $6 THEN 'retrieval'
-          WHEN name = $7 THEN 'data_retention_poll'
```
SgtPooki, line 218: shouldn't we keep these if the jobs are still listed in the table? so our job metrics are accurate?

apps/backend/src/jobs/repositories/job-schedule.repository.ts
```diff
@@ -251,12 +245,10 @@ export class JobScheduleRepository {
       SELECT
         CASE
           WHEN name = $3 THEN COALESCE(data->>'jobType', 'unknown')
-          WHEN name = $4 THEN 'metrics'
-          WHEN name = $5 THEN 'metrics_cleanup'
```
SgtPooki, line 255: ditto

apps/backend/src/jobs/jobs.service.ts
```diff
@@ -1072,6 +996,22 @@ export class JobsService implements OnModuleInit, OnApplicationShutdown {
...
 
       for (const row of rows) {
+        // Skip legacy job types that are no longer supported
+        if (row.job_type === "metrics" || row.job_type === "metrics_cleanup") {
+          this.logger.warn({
+            event: "legacy_job_type_skipped",
+            message: "Skipping legacy job type - please remove from job_schedule_state table",
+            jobType: row.job_type,
+            scheduleId: row.id,
+          });
```
SgtPooki: we should have a follow up issue to remove this code and ensure the rows are all deleted

SgtPooki (approved):
lgtm, a few comments

apps/backend/src/jobs/jobs.service.ts
silent-cipher: Created new issue - #457

apps/backend/src/jobs/repositories/job-schedule.repository.ts
silent-cipher: reverted back in [f012d74](https://github.com/FilOzone/dealbot/pull/436/commits/f012d74d1df5a659216cee7e701a8dc448e63de2)

apps/backend/src/database/entities/job-schedule-state.entity.ts
silent-cipher: Addressed in both of these - [f012d74](https://github.com/FilOzone/dealbot/pull/436/commits/f012d74d1df5a659216cee7e701a8dc448e63de2), [74e93bc](https://github.com/FilOzone/dealbot/pull/436/commits/74e93bc60e8ebe23e88796fd96847104b5278378)
Also, these job types will be removed in the issue #457

Issue #427, opened 2026-04-02
New check: retrieval++


BigLep:
## Done Criteria
We have a retrieval check that does not easily reveal the checker as dealbot by:
1. Running on generic/dynamic infrastructure (rather than just coming from static dealbot k8s clusters).
2. Checking for retrieval of pieces beyond what dealbot created. Anything on chain with an `ipfsRootCid` is fair game.

Once a piece for checking has been determined, it should follow similar behavior to the [basic retrieval check](https://github.com/FilOzone/dealbot/blob/main/docs/checks/retrievals.md) of doing CID looking in IPNI and doing `/ipfs` retrieval. 

All metrics should have [similar dimensionality](https://github.com/FilOzone/dealbot/blob/main/docs/checks/retrievals.md) to [basic retrieval check](https://github.com/FilOzone/dealbot/blob/main/docs/checks/retrievals.md) (e.g., provider info, checkType). A new check type should be used to disambiguate this from the basic retrieval check.

### Open questions
1. What to call this new `checkType` (e.g., `retrievalAnon`)?
2. Piece selection logic/method (e.g., query subgraph X for all pieces with `ipfsRootCid` on provider X)
3. How much `/ipfs` retrieval do we do (e.g., all blocks, n blocks, n MB)?
4. Do we make these "anonymous retrieval checks" form all locations to all SPs or be more geographically aware (e.g., check on EU SPs from the EU). We should follow suit or learn from https://github.com/FilOzone/dealbot/issues/246

## Why Important
This gives a more honest analysis of how an SP is doing. This methodology makes it harder for the SP to be on their best behavior because "the teacher is watching".  

## User/Customer
Prospective/current FOC users who want to have confidence that data stored with FOC will be truly retrievable later by anyone.

## Subtasks

- [x] Anything listed as a sub-issue
- [x] Move the subgraph to the FOC WG goldsky team as discussed in https://github.com/FilOzone/dealbot/issues/498
- [x] Enable on staging
- [x] Enable on production
- [x] Monitoring of the subgraph per https://github.com/FilOzone/dealbot/issues/498
- [ ] (probelab) Add to https://probelab.io/filecoin/foc
- [X] Add to FOC WG BetterStack alarms: https://telemetry.betterstack.com/team/t468215/dashboards/618457/charts/8973679935?rf=now-30d&rt=now

rvagg:
Things we care about:

* Piece retrieval works for whole piece and the bytes are correct
[...]

Issue #246, opened 2026-02-10
Support checks from multiple regions


BigLep:
## Done Criteria
Our core checks (data storage, retrieval) are:
1. Run from multiple regions
2. Have their metrics attributed with the "testing-from" region in our metric datastore
3. Any UI of our metrics make clear where the check was run from 

## Why Important
Dealbot gives a truer sample of performance from wherever it is running.  It's currently running from the EU, which biases its results for EU-based users. 

## Notes
1. This is a placeholder for now.  Given it's not a GA priority, it hasn't been scoped out more.

BigLep:
Dropping down some thoughts on this given its priority goes up if we rely on it as part of https://github.com/FilOzone/dealbot/issues/427

## Geographic identifiers
For public metrics, I think we should is IATA airport codes because:
1. City-level granular - Country is not specific enough, at least for places like the US.
2. Infrastructure-agnostic - We're not encoding the names of a given infra provider like Hetzner / AWS.
3. Globally standardized
4. (According to Claude) Already the de facto convention used by Cloudflare, Fastly, and other major infrastructure providers for exactly this purpose.  (I know internally that AWS uses these codes.)
5. Human-readable

I think it's also good for metrics we view internally for their to be extra details about the infra provider and infra region (e.g., aws:us-east-1, hetzner:fsn1)

## Regional Distribution of Work
I think it's most useful to have "NA/EU/APAC" SPs get tested by infrastructure in "NA/EU/APAC" (i.e., each SP gets tested from one region, not all regions). 

Why?
1. That will give most representative performance numbers, assuming a client uses an SP in their region. (This isn't a valid default assumption though per below)
2. Cleaner dealbot succes signal - We expect it should reduce some of the intermittent failures we have seeing going cross region which has also prompted increasing timeouts ([example](https://github.com/FilOzone/dealbot/issues/383)). 
3. Cleaner dealbot latency/bandwidth signal - Rather than mixing latencies/bandwidths from multiple regions which will inherently have have different performance profiles, we get a read from a single more proximate region.

Why not do this?
1. Synapse (per [private copilot chat](https://github.com/copilot/c/5a202954-08b6-4cf9-8f97-7f41d34d9229) pasted below) doesn't currently do any work to try an select a geographically close SP.  Thus dealbot will make things look better than they actually are in the wild.

I am still in favor of doing this though for moving us in the right direction. If we do, the key things we'll need to adjust in dealbot are:
[...]

PR #438, opened 2026-04-08
feat: write event data to clickhouse 

### LINKED ISSUE #426: Provide hook to write high dimensionality data to a custom clickhouse isntance
## Done Criteria
A clikchouse connection string can be provided and dealbot will log all of its check data with lots of dimensions into that clickhouse database.

## Why Important
Allows for rich analaysis and visualization rather than being limited by prometheus constraints. 
 
## User/Customer
We expect to use this functionality for probelab to be able to get rich dealbot data so they can do analysis/visualization.

## Notes

iand:
Adds a ClickHouse integration to the backend that captures individual deal, retrieval, and data retention observations as queryable analytical records

Three tables are written to: data_storage_checks, retrieval_checks, and data_retention_challenges. Writes are buffered and flushed in the background; failures are logged and swallowed so ClickHouse issues never affect job execution. 

When CLICKHOUSE_URL is unset the integration is a no-op, so the app works unchanged without it configured. Tables are created automatically on startup. A ClickHouse instance is included in the local Kind cluster for development.

Four Prometheus metrics expose ClickHouse health: flush duration, error count, buffer depth, rows inserted.

For #426


apps/backend/src/clickhouse/clickhouse.config.ts
```diff
@@ -0,0 +1,20 @@
...
+  flushIntervalMs: number;
+  probeLocation: string;
+}
+
+export function loadClickhouseConfig(): IClickhouseConfig {
+  return {
+    url: process.env.CLICKHOUSE_URL || undefined,
+    batchSize: Number.parseInt(process.env.CLICKHOUSE_BATCH_SIZE || "500", 10),
+    flushIntervalMs: Number.parseInt(process.env.CLICKHOUSE_FLUSH_INTERVAL_MS || "5000", 10),
+    probeLocation: process.env.DEALBOT_PROBE_LOCATION || "unknown",
```
dennis-tra: With https://github.com/FilOzone/dealbot/issues/427#issuecomment-4205374457 the probe location might be dynamic. If the `backend` will continue to perform probes, I guess the `probeLocation` should be part of the general backend config than the clickhouse config.

apps/backend/src/clickhouse/clickhouse.schema.ts
```diff
@@ -0,0 +1,81 @@
...
+    pieces_confirmed_at         Nullable(DateTime64(3, 'UTC')),
+
+    ipni_status                 LowCardinality(Nullable(String)),
+    ipni_indexed_at             Nullable(DateTime64(3, 'UTC')),
+    ipni_advertised_at          Nullable(DateTime64(3, 'UTC')),
+    ipni_verified_at            Nullable(DateTime64(3, 'UTC')),
+    ipni_verified_cids_count    Nullable(UInt32),
+    ipni_unverified_cids_count  Nullable(UInt32)
+) ENGINE MergeTree()
+  PRIMARY KEY (network, probe_location, sp_address, timestamp)
```
dennis-tra: I know we discussed this schema but thinking about it. I guess we'll have completely different deployments for different networks. Each deployment will probably write its result into a dedicated database. Therefore, I'd probably remove the `network` from the primary key again.

apps/backend/src/clickhouse/clickhouse.service.ts
```diff
@@ -0,0 +1,141 @@
...
+    this.bufferRows.set(this.buffer.length);
+
+    if (this.buffer.length >= this.config.batchSize) {
+      this.flush().catch((err) => {
+        this.logger.error({ event: "flush_batch_error", error: String(err) });
+      });
+    }
+  }
+
+  private async flush(): Promise<void> {
```
dennis-tra, line 108: just something I noticed: calls to `flush` in this file don't provide backpressure. If clickhouse is very slow to respond we would repeatedly call flush where each call may overlap the previous one and potentially amplify the problem. The volume of data that expect won't make a difference I suppose so I don't think anything needs to change. Just wanted to share my thoughts 👍

apps/backend/src/clickhouse/clickhouse.schema.ts
iand: That's a good point, I'll look at that

apps/backend/src/clickhouse/clickhouse.service.ts
iand: Agree with that assessment. One scenario is if clickhouse is unavailable for a period. The buffer is unbounded and held in memory so a prolonged outage would grow it indefinitely but at dealbots expected data rates it would take several days for this to become a problem. If that was a concern then we could implement a circular buffer and accept that there will be data loss if the database is unavailable for long periods.

apps/backend/src/clickhouse/clickhouse.config.ts
iand: Agree. Just made that change

apps/backend/src/clickhouse/clickhouse.schema.ts
iand: Done in #[d5edab1](https://github.com/FilOzone/dealbot/pull/438/commits/d5edab1e249c58f808cf008209de63800fb706bb)

apps/backend/src/clickhouse/clickhouse.config.ts
iand: Done in #0444a3f

iand:
I will look at the copilot issues on Monday

apps/backend/src/clickhouse/clickhouse.service.ts
SgtPooki: noted. good catch

apps/backend/src/clickhouse/clickhouse.service.ts
```diff
@@ -0,0 +1,144 @@
...
+    }, this.config.flushIntervalMs);
+
+    this.logger.log({
+      event: "clickhouse_initialized",
+      host: parsedUrl.host,
+      database: parsedUrl.pathname.replace(/^\//, ""),
+      batchSize: this.config.batchSize,
+      flushIntervalMs: this.config.flushIntervalMs,
+      probeLocation: this.configService.get("app").probeLocation,
+    });
```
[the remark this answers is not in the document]
SgtPooki: this is a good one to fix, but failing worker/startup should be fine and will prevent rollout of the changes without disrupting deployed instances.. however, transient failures later would need this fix to prevent new deployment after an unrelated change.

apps/backend/src/clickhouse/clickhouse.service.ts
```diff
@@ -0,0 +1,144 @@
...
+    // Group by table so we can do one insert call per table
+    const byTable = new Map<string, Record<string, unknown>[]>();
+    for (const { table, row } of batch) {
+      let rows = byTable.get(table);
+      if (!rows) {
+        rows = [];
+        byTable.set(table, rows);
+      }
+      rows.push(row);
+    }
```
SgtPooki, line 123: we never add rows back to `byTable` ?

apps/backend/src/clickhouse/clickhouse.service.ts
```diff
@@ -0,0 +1,144 @@
...
+            format: "JSONEachRow",
+          });
+          this.rowsInserted.inc({ table }, rows.length);
+        }),
+      );
+      end();
+    } catch (err) {
+      end();
+      this.flushErrors.inc();
+      this.logger.error({
```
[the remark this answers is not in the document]
SgtPooki: this review comment seems to be misplaced?

apps/backend/src/data-retention/data-retention.service.ts
```diff
@@ -329,6 +331,15 @@ export class DataRetentionService {
 
     const confirmedTotalSuccess = totalProvingPeriods - totalFaultedPeriods;
 
+    this.clickhouseService.insert("data_retention_challenges", {
+      timestamp: Date.now(),
+      probe_location: this.clickhouseService.probeLocation,
+      sp_address: address,
```
SgtPooki, line 340: I think our PMs and others have mentioned wanting to view SPs by ID instead of by address for simplicity, but i'm not sure if we need that since we are emitting names.. might be worth swapping out if not too much effort?

apps/backend/src/data-retention/data-retention.service.ts
```diff
@@ -329,6 +331,15 @@ export class DataRetentionService {
...
+      sp_name: pdpProvider.name ?? null,
+      total_proving_periods: Number(totalProvingPeriods),
+      total_faulted_periods: Number(totalFaultedPeriods),
+    });
```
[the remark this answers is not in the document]
SgtPooki: comment seems misplaced...

apps/backend/src/deal/deal.service.ts
```diff
@@ -336,6 +338,7 @@ export class DealService implements OnModuleInit, OnModuleDestroy {
               });
               deal.piecesConfirmedTime = new Date();
               deal.status = DealStatus.PIECE_CONFIRMED;
+              deal.pieceId = event.data.pieceIds?.[0] != null ? Number(event.data.pieceIds[0]) : undefined;
```
SgtPooki: not sure if this is needed? did you run into an issue?

apps/backend/src/deal/deal.service.ts
```diff
@@ -732,10 +735,35 @@ export class DealService implements OnModuleInit, OnModuleDestroy {
     // Only set pieceSize here if it hasn't been set earlier in the deal flow.
     deal.pieceSize = pieceSize;
 
-    deal.pieceId = uploadResult.pieceId;
+    deal.pieceId ??= uploadResult.pieceId;
```
SgtPooki: not sure if this is needed either, we should always have pieceId from uploadResult and don't need it before.. but it's been a week or two since i've been in the codebase.

apps/backend/src/deal/deal.service.ts
```diff
@@ -732,10 +735,35 @@ export class DealService implements OnModuleInit, OnModuleDestroy {
...
+      file_size_bytes: deal.fileSize ?? null,
+      piece_size_bytes: deal.pieceSize ?? null,
+      status: deal.status,
+      error_code: deal.errorCode ?? null,
+      retry_count: deal.retryCount,
+      upload_started_at: deal.uploadStartTime?.getTime() ?? null,
+      upload_ended_at: deal.uploadEndTime?.getTime() ?? null,
+      pieces_added_at: deal.piecesAddedTime?.getTime() ?? null,
+      pieces_confirmed_at: deal.piecesConfirmedTime?.getTime() ?? null,
+      ipni_status: deal.ipniStatus ?? null,
```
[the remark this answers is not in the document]
SgtPooki: i think this is misplaced, but worth checking into

apps/backend/src/deal/deal.service.ts
```diff
@@ -732,10 +735,35 @@ export class DealService implements OnModuleInit, OnModuleDestroy {
...
+      sp_address: deal.spAddress,
+      sp_name: deal.storageProvider?.name ?? null,
+      deal_id: deal.id,
+      piece_cid: deal.pieceCid ?? null,
+      piece_id: deal.pieceId ?? null,
+      file_size_bytes: deal.fileSize ?? null,
+      piece_size_bytes: deal.pieceSize ?? null,
+      status: deal.status,
+      error_code: deal.errorCode ?? null,
+      retry_count: deal.retryCount,
```
SgtPooki: idk if this retryCount is useful at all..  the only retries i know of is ipni checking.. maybe its an old property that hasn't been stripped out yet?

each `job` runs a `check` only once.. if it succeeds/fails, we just emit that it did.

apps/backend/src/retrieval/retrieval.service.ts
```diff
@@ -255,7 +257,25 @@ export class RetrievalService {
       this.retrievalMetrics.recordHttpResponseCode(providerLabels, executionResult.metrics.statusCode);
     }
 
-    return this.saveRetrieval(retrieval);
+    const saved = await this.saveRetrieval(retrieval);
+
+    this.clickhouseService.insert("retrieval_checks", {
+      timestamp: Date.now(),
+      probe_location: this.clickhouseService.probeLocation,
+      sp_address: deal.spAddress,
```
SgtPooki, line 265: maybe sp_id as well?

kustomize/overlays/local/clickhouse/deployment.yaml
```diff
@@ -0,0 +1,61 @@
...
+kind: Deployment
+metadata:
+  name: dealbot-clickhouse
+  labels:
+    app.kubernetes.io/name: dealbot-clickhouse
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: dealbot-clickhouse
```
SgtPooki, line 11: I love that you set up a local overlay so we can test this.. gonna try this in a bit.

Makefile
```diff
@@ -119,6 +120,15 @@ redeploy:
...
 	$(MAKE) restart
 
+# Delete the ClickHouse PVC so initdb scripts run again on next deploy.
+# Use this when the ClickHouse schema has changed and you need a clean state.
+clickhouse-reset:
+	-kubectl delete pvc -n $(NAMESPACE) dealbot-clickhouse
+	-kubectl delete pod -n $(NAMESPACE) -l app.kubernetes.io/name=dealbot-clickhouse
+
+clickhouse-shell:
+	kubectl exec -it -n $(NAMESPACE) deployment/dealbot-clickhouse -- clickhouse-client --database dealbot
```
SgtPooki, line 130: 😍

apps/backend/src/clickhouse/clickhouse.service.ts
```diff
@@ -0,0 +1,144 @@
...
+      clearInterval(this.flushTimer);
+      this.flushTimer = null;
+    }
+    await this.flush();
+    await this.client?.close();
+  }
+
+  /**
+   * Queue a row for insertion. Returns immediately; the flush happens in the background.
+   * Safe to call when ClickHouse is disable: rows are silently dropped.
```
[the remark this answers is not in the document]
SgtPooki: should fix

apps/backend/src/clickhouse/clickhouse.service.ts
```diff
@@ -0,0 +1,144 @@
...
+      host: parsedUrl.host,
+      database: parsedUrl.pathname.replace(/^\//, ""),
+      batchSize: this.config.batchSize,
+      flushIntervalMs: this.config.flushIntervalMs,
+      probeLocation: this.configService.get("app").probeLocation,
+    });
+  }
+
+  private async migrate(database: string): Promise<void> {
+    const migrations = buildMigrations(database);
```
SgtPooki, line 70: we should guard this on this.client existence as well, and remove the `this.client!` non-null assertion

apps/backend/src/clickhouse/clickhouse.service.ts
```diff
@@ -0,0 +1,144 @@
...
+      return;
+    }
+
+    this.client = createClient({
+      url: this.config.url,
+    });
+
+    const parsedUrl = new URL(this.config.url!);
+    await this.migrate(parsedUrl.pathname.replace(/^\//, ""));
+
```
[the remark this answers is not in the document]
SgtPooki: this is a very bad comment. lol. i'm resolving.

apps/backend/src/clickhouse/clickhouse.service.ts
```diff
@@ -0,0 +1,144 @@
...
+    if (!this.config.url) {
+      this.logger.log("CLICKHOUSE_URL not set, writes to ClickHouse disabled");
+      return;
+    }
+
+    this.client = createClient({
+      url: this.config.url,
+    });
+
+    const parsedUrl = new URL(this.config.url!);
```
SgtPooki: we should not use non-null assertions, and this should not be possible to be null with the guard clause at function entry.

apps/backend/src/clickhouse/clickhouse.schema.ts
```diff
@@ -0,0 +1,79 @@
+/**
+ * ClickHouse DDL statements executed on startup via CREATE DATABASE/TABLE IF NOT EXISTS.
+ * Order matters: database must be created before tables.
+ */
+export function buildMigrations(database: string): string[] {
+  return [
```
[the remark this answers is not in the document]
SgtPooki: I'm assuming we're going to get some clickhouse URLs from probelab that are conigured with specific database?

apps/backend/src/clickhouse/clickhouse.config.ts
```diff
@@ -0,0 +1,18 @@
...
+  flushIntervalMs: number;
+}
+
+export function loadClickhouseConfig(): IClickhouseConfig {
+  return {
+    url: process.env.CLICKHOUSE_URL || undefined,
+    batchSize: Number.parseInt(process.env.CLICKHOUSE_BATCH_SIZE || "500", 10),
+    flushIntervalMs: Number.parseInt(process.env.CLICKHOUSE_FLUSH_INTERVAL_MS || "5000", 10),
+  };
+}
```
SgtPooki: do we need this? we have app config where a lot of this is duplicated, right?

SgtPooki (changes requested):
a few changes requested, a lot of general comments and responses to copilot things, I want to run the local cluster and see how things fare.

apps/backend/src/deal/deal.service.ts
silent-cipher: I had raised a PR for these same changes yesterday (#439).
This seems important since uploadResult often misses pieceId, and these changes help preserve it.

SgtPooki:
ran this locally and it seems to be working fine, but I would like to see the try/catch addressed. I might get to this before you do, we'll see.. 

I did add https://github.com/FilOzone/dealbot/pull/443 to address some errors i saw during startup, but not blocking at all

apps/backend/src/clickhouse/clickhouse.schema.ts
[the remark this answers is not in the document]
BigLep: Yeah, that's right. Probelab has a clickhouse account they use and will set up new DB for Dealbot data. (FOC WG could create a clickhouse if needed/wanted in future.)

apps/backend/src/clickhouse/clickhouse.schema.ts
```diff
@@ -0,0 +1,79 @@
...
+    `CREATE TABLE IF NOT EXISTS ${database}.data_storage_checks
+(
+    timestamp                   DateTime64(3, 'UTC'),
+
+    probe_location              LowCardinality(String),
```
BigLep: Per some of the proposal in https://github.com/FilOzone/dealbot/issues/246, I assume we'd do something like `hetzner:fsn1`, `aws:us-east-1`)

Basically I think we want to know where the probe was run from on which infrastructure.

apps/backend/src/clickhouse/clickhouse.schema.ts
```diff
@@ -0,0 +1,79 @@
...
+    sp_address                  String,
+    sp_name                     Nullable(String),
+
+    deal_id                     UUID,
+    piece_cid                   Nullable(String),
+    piece_id                    Nullable(UInt64),
+    file_size_bytes             Nullable(UInt64),
```
BigLep: Can we bring in the comments like were in https://github.com/FilOzone/dealbot/issues/426 "raw file size before CAR encoding"

apps/backend/src/clickhouse/clickhouse.schema.ts
```diff
@@ -0,0 +1,79 @@
...
+    `CREATE TABLE IF NOT EXISTS ${database}.retrieval_checks
+(
+    timestamp               DateTime64(3, 'UTC'),
+    probe_location          LowCardinality(String),
+    sp_address              String,
+    sp_name                 Nullable(String),
+
+    deal_id                 Nullable(UUID),
+    retrieval_id            UUID,
+    service_type            LowCardinality(String),
```
BigLep: I would want a comment on this

apps/backend/src/clickhouse/clickhouse.schema.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    status                  LowCardinality(String),
+    http_response_code      Nullable(UInt16),
+    retry_count             UInt8 DEFAULT 0,
+
+    ttfb_ms                 Nullable(Float64),
+    last_byte_ms            Nullable(Float64),
```
BigLep: I'd like more symmetry in the names

In https://github.com/FilOzone/dealbot/blob/main/docs/checks/events-and-metrics.md#time-related-metrics we did:

ipfsRetrievalFirstByteMs
ipfsRetrievalLastByteMs

I generally think it's good to align with that doc, but if we want to be more concise

first_byte_ms
last_byte_ms

?

apps/backend/src/clickhouse/clickhouse.schema.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    `CREATE TABLE IF NOT EXISTS ${database}.data_retention_challenges
+(
+    timestamp               DateTime64(3, 'UTC'),
+    probe_location          LowCardinality(String),
+    sp_address              String,
+    sp_name                 Nullable(String),
+
+    total_proving_periods   UInt32,
+    total_faulted_periods   UInt32
```
BigLep: This likely needs some more clarity.
We want to know about challenges (how many succeeded vs. failed)
We also need to account for overdue periods , which we'll treat as a failure.
I'm realizing that https://github.com/FilOzone/dealbot/blob/main/docs/checks/events-and-metrics.md#time-related-metrics isn't enumerating this correctly/fully.

@silent-cipher : can you please shed some light here on specifics and also make sure we get https://github.com/FilOzone/dealbot/blob/main/docs/checks/events-and-metrics.md#time-related-metrics updated to account for all the count metrics?

BigLep (commented):
A few comments when looking at the schema. Will leave to engineers to give the approval on the PR.

apps/backend/src/clickhouse/clickhouse.schema.ts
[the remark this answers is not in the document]
iand: Yes, the databases will be created externally. Dealbot then owns its schema within that database,

apps/backend/src/clickhouse/clickhouse.schema.ts
iand: Yes, will carry those comments over

apps/backend/src/clickhouse/clickhouse.schema.ts
iand: changed to `first_byte_ms` for symmetry with `last_byte_ms`

apps/backend/src/data-retention/data-retention.service.ts
iand: Added sp_id to schema

apps/backend/src/deal/deal.service.ts
iand: Yeah, this was often null

apps/backend/src/deal/deal.service.ts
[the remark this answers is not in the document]
iand: For DateTime64(3), ClickHouse interprets numeric inputs as ticks scaled to the precision. With precision 3, a number is read as milliseconds since epoch.

apps/backend/src/retrieval/retrieval.service.ts
```diff
@@ -255,7 +257,25 @@ export class RetrievalService {
...
+      deal_id: deal.id,
+      retrieval_id: saved.id,
+      service_type: saved.serviceType,
+      status: saved.status,
+      http_response_code: executionResult.metrics.statusCode || null,
+      retry_count: saved.retryCount,
+      ttfb_ms: executionResult.success ? executionResult.metrics.ttfb : null,
+      last_byte_ms: executionResult.success ? executionResult.metrics.latency : null,
+      bytes_retrieved: executionResult.success ? executionResult.metrics.responseSize : null,
+    });
```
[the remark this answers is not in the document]
iand: For DateTime64(3), ClickHouse interprets numeric inputs as ticks scaled to the precision. With precision 3, a number is read as milliseconds since epoch.

apps/backend/src/clickhouse/clickhouse.schema.ts
```diff
@@ -0,0 +1,79 @@
...
+    ipni_status                 LowCardinality(Nullable(String)),
+    ipni_indexed_at             Nullable(DateTime64(3, 'UTC')),
+    ipni_advertised_at          Nullable(DateTime64(3, 'UTC')),
+    ipni_verified_at            Nullable(DateTime64(3, 'UTC')),
+    ipni_verified_cids_count    Nullable(UInt32),
+    ipni_unverified_cids_count  Nullable(UInt32)
+) ENGINE MergeTree()
+  PRIMARY KEY (probe_location, sp_address, timestamp)
+  PARTITION BY toStartOfMonth(timestamp)
+  TTL toDateTime(timestamp) + INTERVAL 1 YEAR`,
```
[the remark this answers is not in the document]
iand: ClickHouse documentation states that if PRIMARY KEY is specified without ORDER BY, the sort key is set equal to the primary key. This has been the case for a long time, so any comment about ClickHouse version seems irrelevant. Besides, with the small amounts of data we are dealing with the ordering is not going to have any real effect on query times.

apps/backend/src/clickhouse/clickhouse.schema.ts
iand: added comments in #357c099

apps/backend/src/clickhouse/clickhouse.service.ts
iand: Done

apps/backend/src/clickhouse/clickhouse.service.ts
iand: Done

apps/backend/src/clickhouse/clickhouse.service.ts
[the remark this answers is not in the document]
iand: Done

apps/backend/src/clickhouse/clickhouse.service.ts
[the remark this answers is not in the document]
iand: Agree, I don't see any change to make

apps/backend/src/data-retention/data-retention.service.ts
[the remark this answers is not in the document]
iand: See other replies about the DateTime64(3) resolution

apps/backend/src/deal/deal.service.ts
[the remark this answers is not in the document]
iand: So, no change needed

apps/backend/src/retrieval/retrieval.service.ts
iand: Added

apps/backend/src/retrieval/retrieval.service.ts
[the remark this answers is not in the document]
iand: No change needed

apps/backend/src/clickhouse/clickhouse.schema.ts
iand: Added in #357c099

apps/backend/src/clickhouse/clickhouse.schema.ts
iand: Yes, we'll have a standard set of locations and these are set with an environment variable in the dealbot deploy, then carried through to the database

apps/backend/src/clickhouse/clickhouse.config.ts
iand: I just split it to keep self contained, but will combine with the rest of the config

iand: Moved to app config

apps/backend/src/clickhouse/clickhouse.service.ts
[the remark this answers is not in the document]
iand: What do you want to do here, in light of #443? 

My view is that if we have configured a CLICKHOUSE_URL then we always want to fail at this point. We don't want to continue without clickhouse since we'd be dropping data silently. If the DDL is invalid then we can't write to clickhouse anyway if we managed to run.

Added try/catch to log any error and then fail.  Any error will propagate and fail module initialisation causing the pod to enter crash loop backoff until clickhouse is available.

apps/backend/src/clickhouse/clickhouse.service.ts
iand: rows holds a reference to the array in byTable, rows.push(row) mutates the referenced array. At least, that it is my understanding - I'm not a typescript expert by any measure

apps/backend/src/clickhouse/clickhouse.schema.ts
iand: We could update the schema to:

```sql
  total_periods_due        UInt32,  -- cumulative proving periods due (confirmed by subgraph)                                                             
  total_faulted_periods    UInt32,  -- cumulative periods where proof was not submitted                                                                   
  total_success_periods    UInt32,  -- cumulative periods where proof was submitted (= due - faulted)                                                     
  estimated_overdue_periods UInt32  -- estimated periods not yet recorded on-chain but past deadline  
```
estimated_overdue_periods is computed on lines 325–330

Alternatively, if we want to spend more time thinking about this we could drop this entire table from this PR and add it at a later date. The existing migration code will just handle the addition when we add it.

iand:
I think I have resolved all the straightforward issues. 

- Added comments to the schema.
- Renamed ttfb_ms column to first_byte_ms
- Add sp_id columns to record the storage provider id
- Made stylistic changes for guarding against undefined variables
- Moved clickhouse config into app config for consistency

There were some review comments by copilot that were erroneous, probably due to clickhouse specifics:

- Datetime(3) resolution - the 3 indicates millisecond resolution which is what we want
- ORDER BY on tables - not needed in any recent version of clickhouse

Some comments that need more discussion/response:

1. migrate() errors in onModuleInit - I added try/catch to log error then rethrow it. Are you ok with this approach @SgtPooki
2. add rows back to byTable - I think it's ok as-is @SgtPooki
3. retry_count - I left this in because it could be useful, but it's marginal
4. data_retention_challenges schema - made a suggestion for schema change @BigLep

apps/backend/src/clickhouse/clickhouse.service.ts
```diff
@@ -0,0 +1,150 @@
...
+            table,
+            values: rows,
+            format: "JSONEachRow",
+          });
+          this.rowsInserted.inc({ table }, rows.length);
+        }),
+      );
+      end();
+    } catch (err) {
+      end();
```
SgtPooki: this `end()` and flushErrors should be in a finally block, wrapped in try/catch

apps/backend/src/clickhouse/clickhouse.service.ts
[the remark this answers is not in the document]
SgtPooki: i'm good with that fix. thanks

apps/backend/src/clickhouse/clickhouse.service.ts
```diff
@@ -0,0 +1,150 @@
...
+    for (const { table, row } of batch) {
+      let rows = byTable.get(table);
+      if (!rows) {
+        rows = [];
+        byTable.set(table, rows);
+      }
+      rows.push(row);
+    }
+
+    const end = this.flushDuration.startTimer();
```
SgtPooki, line 125: clickhouseFlushDurationMs is milliseconds, but startTimer returns seconds: https://github.com/siimon/prom-client#utility-methods

we likely need to change the histogram or use another timer utility

apps/backend/src/clickhouse/clickhouse.service.ts
```diff
@@ -0,0 +1,150 @@
...
+      this.flush().catch((err) => {
+        this.logger.error({ event: "flush_batch_error", error: String(err) });
+      });
+    }
+  }
+
+  private async flush(): Promise<void> {
+    if (!this.client || this.buffer.length === 0) return;
+
+    const batch = this.buffer.splice(0, this.buffer.length);
```
SgtPooki: This drains `this.buffer` before ClickHouse accepts the rows. If any insert below fails, the catch block only logs `droppedRows`, so a transient ClickHouse/network failure permanently loses the whole batch. Should we either requeue failed rows for a later flush, or make the intentional drop policy explicit and bounded?

apps/backend/src/clickhouse/clickhouse.service.ts
SgtPooki: you're right, sorry for the churn

apps/backend/src/data-retention/data-retention.service.ts
```diff
@@ -332,6 +334,16 @@ export class DataRetentionService {
 
     const confirmedTotalSuccess = totalProvingPeriods - totalFaultedPeriods;
 
+    this.clickhouseService.insert("data_retention_challenges", {
```
SgtPooki, line 337: we compute `estimatedOverduePeriods` above and emit it to Prometheus later, but the ClickHouse row only stores cumulative proving/faulted totals. Since overdue periods are part of how we reason about data-retention failures, should this table also include `estimated_overdue_periods` (or should we defer the retention table until the schema is settled)?

SgtPooki (changes requested):

apps/backend/src/deal/deal.service.ts
SgtPooki: @iand we probably don't need retry count

apps/backend/src/clickhouse/clickhouse.service.ts
iand: I don't think `flushErrors.Inc()` should be in the finally block, since it would then increment on every flush, not just on errors.

I have moved `end()` into a finally block.

apps/backend/src/clickhouse/clickhouse.service.ts
iand: Renamed to clickhouseFlushDurationSeconds so we can continue to use startTimer

apps/backend/src/clickhouse/clickhouse.service.ts
iand: Good catch. Changed this to take a slice of the buffer first, then clear the buffer only on successful insert. This keeps the rows in the buffer in the event of clickhouse unavailability.

I also added a max buffer size config variable and changed insert to treat the buffer as circular so we bound the number of rows in the buffer at a constant level. If clickhouse is unavailable for a long period then we end up with a gap in the data, but this is acceptable for telemetry like this.

apps/backend/src/data-retention/data-retention.service.ts
iand: I think it should include `estimated_overdue_periods`. See my earlier comment to steve about a revised schema

iand:
@SgtPooki I've hopefully addressed your latest comments

apps/backend/src/clickhouse/clickhouse.schema.ts
SgtPooki: i am good with your recommended approach and think we can adjust as necessary so we can unblock you @iand

SgtPooki (changes requested):
@iand the only thing left is to move forward with your schema updates to get you unblocked (see https://github.com/FilOzone/dealbot/pull/438#discussion_r3069780112), and also the retry_count can likely be dropped (see https://github.com/FilOzone/dealbot/pull/438#discussion_r3120049081)

iand:
Thanks @SgtPooki, I removed the retry_count columns as suggested.

Final decision is needed by @BigLep on the `data_retention_challenges` schema

apps/backend/src/clickhouse/clickhouse.schema.ts
BigLep: Agreed. Your proposal sounds good @iand - thanks

iand:
All changes from review comments are addressed now

apps/backend/src/clickhouse/clickhouse.schema.ts
silent-cipher: Raised a pr to update docs to clarify between challenges and periods - https://github.com/FilOzone/dealbot/pull/476

iand:
Is there anything remaining blocking this from being approved and merged @SgtPooki ?

iand:
Actually, one change we could make to the `retrieval_checks` table is to add a `retrieval_type` column

```SQL
retrieval_type            LowCardinality(String),  -- 'deal' | 'sample'
```

The two values are:

- `deal` - a retrieval triggered from a known deal id
- `sample` - a retrieval triggered from sampling a random piece, in this case deal_id will be null

Not I used `sample` here rather than `anonymous` as used in https://github.com/FilOzone/dealbot/issues/427 and https://github.com/FilOzone/dealbot/pull/459 since I think it better conveys the concept, but we could align on `anon` if preferred

/cc @BigLep  @dennis-tra @SgtPooki

SgtPooki:
> Not I used sample here rather than anonymous as used in https://github.com/FilOzone/dealbot/issues/427 and https://github.com/FilOzone/dealbot/pull/459 since I think it better conveys the concept, but we could align on anon if preferred

I'm good with either sample or anon, we just need to make sure we're using the same language.

SgtPooki (approved):

PR #439, opened 2026-04-09
fix: preserve pieceId during deal creation

silent-cipher:
## Summary 

These changes address `pieceId` preservation issues. `uploadResult` is often missing `pieceId`, so we preserve it from the `piecesConfirmed` event. This is important for cleaning up pieces later on.

## Problem 

Missing `pieceId` while testing locally - 
```
Select count(*) from deals where status = 'deal_created' and piece_id is null;
 count 
-------
    25
(1 row)

Select count(*) from deals where status = 'deal_created';
 count 
-------
    68
(1 row)
```


apps/backend/src/deal/deal.service.ts
```diff
@@ -334,6 +334,7 @@ export class DealService implements OnModuleInit, OnModuleDestroy {
                 message: "Pieces confirmed",
                 pieceIds: event.data.pieceIds,
               });
+              deal.pieceId = event.data.pieceIds.length > 0 ? Number(event.data.pieceIds[0]) : undefined;
```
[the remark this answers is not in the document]
silent-cipher: `pieceId` values are currently expected to be within `Number.MAX_SAFE_INTEGER`

apps/backend/src/deal/deal.service.ts
```diff
@@ -334,6 +334,9 @@ export class DealService implements OnModuleInit, OnModuleDestroy {
                 message: "Pieces confirmed",
                 pieceIds: event.data.pieceIds,
               });
+              if (event.data.pieceIds.length > 0) {
+                deal.pieceId = Number(event.data.pieceIds[0]);
```
SgtPooki, line 384: i'm pretty sure we should only ever have one pieceId with dealbot content, but we should probably log a warning/error if there is more than one

SgtPooki (approved):

apps/backend/src/deal/deal.service.ts
silent-cipher: [3c3cea1](https://github.com/FilOzone/dealbot/pull/439/commits/3c3cea17295fcf59a76f885af17e99893e4e16b9)

PR #442, opened 2026-04-09
refactor: drop legacy metrics database schema

Chaitu-Tatipamula:
## Summary

Adds a database migration to drop the legacy metrics schema objects and removes the corresponding TypeORM entity files and module registrations. This completes step 9 of [#275](https://github.com/FilOzone/dealbot/issues/275#issuecomment-4069535483).

Depends on #441.

### [NEW] Migration `1762000000000-DropMetricsSchema`

Drops the following database objects in dependency order:
1. `refresh_sp_performance_last_week()` and `refresh_sp_performance_all_time()` functions
2. `sp_performance_last_week` and `sp_performance_all_time` materialized views
3. `metrics_daily` table (with `CASCADE`)
4. `metrics_daily_metric_type_enum` and `metrics_daily_service_type_enum` types

> ⚠️ This migration is irreversible. Restore from backup if rollback is needed.

### Deleted entity files
- `metrics-daily.entity.ts`
- `sp-performance-all-time.entity.ts`
- `sp-performance-last-week.entity.ts`

### Modified
- **`database.module.ts`** — Removed `MetricsDaily`, `SpPerformanceAllTime`, `SpPerformanceLastWeek` from entity registrations, `forFeature`, providers, and exports
- **`types.ts`** — Removed `MetricType` enum (only consumer was the deleted entity)

## Verification
- `pnpm typecheck` ✅
- `pnpm test` ✅ (249 tests)
- `pnpm check:ci` ✅


silent-cipher (changes requested):
Could you update the e2e migration tests to include the new migration `DropMetricsSchema1762000000000`?

Chaitu-Tatipamula:
> Could you update the e2e migration tests to include the new migration `DropMetricsSchema1762000000000`?

 so i was looking into this and there is a subtle issue with including `DropMetricsSchema` in the existing e2e roundtrip test.
the problem is `DropMetricsSchema` is irreversible. the existing roundtrip test undoes 7 migrations from the end to get back to before `RemoveSpReceivedRetrieveRequest`, then re-applies them. But several of those undone migrations (`RemoveIpniRetrievedColumns`, `RemoveSpReceivedRetrieveRequest`) have down() methods that `ALTER TABLE metrics_daily`, which no longer exists after `DropMetricsSchema` irreversibly drops it.
maybe we can use isolated databases or at worst adding `metrics_daily` existence guards to every migration that references it wdyt?

PR #445, opened 2026-04-14
refactor: replace ConsoleLogger with nestjs-pino for ISO 8601 timestamps

silent-cipher:
## Summary

Backend JSON logs emitted a Unix millisecond integer for the timestamp field (`"timestamp":1775081543138`), making logs difficult to read in production tooling without post-processing. This PR replaces the default NestJS `ConsoleLogger` with [nestjs-pino](https://github.com/iamolegga/nestjs-pino) / `pino-http` to produce human-friendly ISO 8601 timestamps while preserving the existing ConsoleLogger-compatible JSON field layout.

### Before:
```json
{"level":"log","pid":12671,"timestamp":1776163682800,"message":{"event":"synapse_initialization","message":"Creating shared Synapse instance"},"context":"DealService"}
{"level":"log","pid":12671,"timestamp":1776163682802,"message":{"event":"pgboss_initialization","message":"Starting pg-boss initialization"},"context":"JobsService"}
```

### After:
```json
{"level":"log","timestamp":"2026-04-14T11:05:01.191Z","pid":13785,"context":"DealService","message":{"event":"synapse_initialization","message":"Creating shared Synapse instance"}}
{"level":"log","timestamp":"2026-04-14T11:05:01.193Z","pid":13785,"context":"JobsService","message":{"event":"pgboss_initialization","message":"Starting pg-boss initialization"}}
```

closes #432


apps/backend/src/common/pino.config.ts
```diff
@@ -0,0 +1,31 @@
...
+};
+
+export function resolvePinoLevel(level: string | undefined): string {
+  if (!level) return "info";
+  return PINO_LEVEL_MAP[level.toLowerCase().trim()] ?? "info";
+}
+
+export function buildLoggerModuleParams() {
+  return {
+    pinoHttp: {
```
rjan90, line 39: This shared `pinoHttp` config leaves `autoLogging` at its default, so wiring `LoggerModule.forRoot() `here will emit an extra "request completed" access log for every route. The previous `ConsoleLogger` setup didn’t do that, so this changes log volume/schema in addition to timestamp formatting.

If the goal is only ISO 8601 timestamps, could we set `autoLogging: false `here?

rjan90 (approved):
One small nit to reduce extra logging. Else it looks good to me

apps/backend/src/common/pino.config.ts
silent-cipher: [2f94322](https://github.com/FilOzone/dealbot/pull/445/commits/2f94322aa0b873126a1fb9ea9d4785c4252afdfc)

apps/backend/src/main.ts
```diff
@@ -48,8 +38,11 @@ async function bootstrap() {
   const app = await NestFactory.create(rootModule, {
     // Allow bootstrap errors to propagate so our catch path can emit structured fatal logs.
     abortOnError: false,
-    logger,
+    // Buffer framework logs during startup; flushed once pino logger is activated below.
+    bufferLogs: true,
```
SgtPooki, line 42: it appears that we need to do this because we're instantiating a different logger than the built in one?

SgtPooki (approved):
lgtm, but i am running locally quick to confirm that logs are appearing properly

SgtPooki:
logs seem to be outputting fine

PR #452, opened 2026-04-15
feat: add sp blocklist to skip checks for specific providers

### LINKED ISSUE #364: dealbot should be able to enable/disable testing against specific SPs
We should be able to block providers for any/all checks by providerId

silent-cipher:
Closes #364


SgtPooki (approved):
in general lgtm, i opened a small pr on top: #455

PR #453, opened 2026-04-16
fix: redact RPC tokens from logged deal errors

rjan90:
## What changed
- redact sensitive credentials from URLs embedded in structured error messages and stacks
- persist a redacted deal failure message instead of the raw upstream error text
- add regression tests covering log serialization and failed deal persistence

## Why
A `429` from the configured Filecoin RPC can bubble up through `viem` with the full request URL in the error message. Dealbot was forwarding that raw message into logs and into `deals.error_message`, which could expose RPC query-string tokens.

## Impact
This keeps the existing error flow intact while replacing values for sensitive URL credentials and query parameters like `token`, `api_key`, `apikey`, `access_token`, and `authorization` with `REDACTED` in the affected dealbot surfaces.

## Root cause
`viem` includes the full request URL in `HttpRequestError` / `ContractFunctionExecutionError` messages, and dealbot was serializing and persisting `error.message` without any redaction.

## Validation
- `pnpm test -- logging.spec.ts deal.service.spec.ts` (backend suite passed)
- `pnpm typecheck`

### DIFF
--- a/apps/backend/src/common/logging.ts
+++ b/apps/backend/src/common/logging.ts
@@ -1,5 +1,6 @@
 type ErrorWithCode = Error & { code?: unknown };
 const MAX_ERROR_STACK_LENGTH = 4 * 1024;
+const ERROR_STACK_REDACTION_BUFFER_LENGTH = 512;
 const MAX_CAUSE_DEPTH = 5;
 const URL_PATTERN = /\bhttps?:\/\/[^\s<>"')\]}]+/gi;
 const REDACTED_VALUE = "REDACTED";
@@ -29,8 +30,11 @@ function redactSensitiveUrl(url: string): string {
   try {
     const parsed = new URL(url);
 
-    if (parsed.username || parsed.password) {
+    if (parsed.username) {
       parsed.username = REDACTED_VALUE;
+    }
+
+    if (parsed.password) {
       parsed.password = REDACTED_VALUE;
     }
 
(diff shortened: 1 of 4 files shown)


apps/backend/src/common/logging.ts
```diff
@@ -21,6 +25,31 @@ export function toJsonSafe(value: unknown): unknown {
...
+
+    for (const key of parsed.searchParams.keys()) {
+      if (SENSITIVE_QUERY_PARAMETER_NAMES.has(key.toLowerCase())) {
+        parsed.searchParams.set(key, REDACTED_VALUE);
+      }
+    }
+
+    return parsed.toString();
+  } catch {
+    return url;
```
silent-cipher: Shouldn't we return placeholder like `INVALID_URL` if `new URL(url)` throws?

rvagg (approved):
sgtm but the copilot suggestion is important to fix before merging

apps/backend/src/common/logging.ts
rvagg: actually pretty interesting if it does fail here, what kind of url are we encountering? maybe a `wss`, perhaps we just remove the `try/catch` entirely and see how it fails, or ```return `INVALID_URL[${url}]` ``` so we don't lose everything

apps/backend/src/common/logging.ts
```diff
@@ -21,6 +25,31 @@ export function toJsonSafe(value: unknown): unknown {
...
+  }
+}
+
+export function redactSensitiveText(text: string): string {
+  return text.replace(URL_PATTERN, redactSensitiveUrl);
```
rvagg:
```suggestion
  return text.replaceAll(URL_PATTERN, redactSensitiveUrl);
```

apps/backend/src/common/logging.ts
silent-cipher: can throw for malformed url with too large port number

rvagg: well, it'll throw when the regex matches something that `URL` can't deal with, so that's the interesting intersection as far as this code is concerned, so let's highlight it for now and see what it yields if anything:

```suggestion
    return `INVALID_URL[${url}]`;
```

apps/backend/src/common/logging.ts
```diff
@@ -21,6 +25,31 @@ export function toJsonSafe(value: unknown): unknown {
   }
 }
 
+function redactSensitiveUrl(url: string): string {
+  try {
+    const parsed = new URL(url);
+
+    if (parsed.username || parsed.password) {
+      parsed.username = REDACTED_VALUE;
```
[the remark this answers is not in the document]
rjan90: https://github.com/FilOzone/dealbot/pull/453/changes/19ab1c3934be9ae98b4925e5850164d4c994e8a7

apps/backend/src/common/logging.ts
```diff
@@ -48,17 +75,17 @@ export function toStructuredError(error: unknown, depth: number = 0): Structured
     return {
       type: "error",
       name: error.name,
-      message: error.message,
+      message: redactSensitiveText(error.message),
       code: stringCode && stringCode.length > 0 ? stringCode : undefined,
-      stack: truncateErrorStack(error.stack),
+      stack: truncateErrorStack(error.stack ? redactSensitiveText(error.stack) : undefined),
```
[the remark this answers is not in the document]
rjan90: https://github.com/FilOzone/dealbot/pull/453/changes/19ab1c3934be9ae98b4925e5850164d4c994e8a7

SgtPooki (approved):

PR #469, opened 2026-04-23
feat: dealbot-specific subgraph

dennis-tra:
This PR is a follow-up of #459 and adds only the subgraph configuration. This was requested by @SgtPooki in [this comment](https://github.com/FilOzone/dealbot/pull/459#issuecomment-4304939784). I've marked the other PR as draft for now.

As mentioned in the other PR this is largely based on the PDP-Explorer but I have stripped everything that's not needed by dealbot. Turns out, a lot wasn't needed.

The other PR has 19 commits with changes to the subgraph scattered among them. Therefore, this PR only has a single commit with the final state of the other PR.

## Subgraph

I have deployed the new subgraphs:

- `mainnet`: https://api.goldsky.com/api/public/project_cmo9sxe5xd4ai01x8cpageyid/subgraphs/dealbot-mainnet/0.3.0/gn
- `calibration`: https://api.goldsky.com/api/public/project_cmo9sxe5xd4ai01x8cpageyid/subgraphs/dealbot-calibration/0.3.0/gn


apps/subgraph/src/pdp-verifier.ts
```diff
@@ -0,0 +1,232 @@
...
+
+  let periodsSkipped: BigInt = BigInt.zero();
+  let faultedPeriods: BigInt = BigInt.zero();
+  let nextDeadline: BigInt;
+
+  if (proofSet.nextDeadline.equals(BigInt.zero())) {
+    // First-init: promote to PROVING, seed maxProvingPeriod.
+    proofSet.status = DataSetStatus.PROVING;
+    proofSet.maxProvingPeriod = BigInt.fromI32(maxProvingPeriodFor(event.address));
+    nextDeadline = currentBlockNumber.plus(proofSet.maxProvingPeriod);
```
SgtPooki: `NextProvingPeriod` includes `challengeEpoch`, which represents the contract's scheduled proof epoch, but this handler derives `nextDeadline` from `event.block.number`. If the `nextProvingPeriod` transaction is mined late, the stored deadline shifts forward relative to the contract's schedule. Because Dealbot identifies overdue sets via `nextDeadline_lt: $blockNumber`, that shift can make overdue sets invisible to the bot until the synthetic deadline catches up. Can we anchor `nextDeadline` to `event.params.challengeEpoch` here? I think it’s also worth checking whether the skipped/faulted-period math should be derived from the emitted epoch rather than `event.block.number`.

basic flow of things without any change:

1. The contract says: “the next proving checkpoint is epoch X.”
1. The subgraph stores: “the next proving checkpoint is when the tx landed, plus one period.”
1. If the tx landed late, those are different.
1. Dealbot trusts the subgraph deadline.
1. So Dealbot can think “not overdue yet” even when the contract schedule says it already is.

apps/subgraph/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "@dealbot/subgraph",
+  "private": true,
+  "license": "(MIT OR Apache-2.0)",
+  "scripts": {
+    "codegen": "graph codegen",
```
SgtPooki: Can we source these ABIs from the canonical Filecoin services/Synapse package instead of committing hand-copied ABI JSON in here? The repo already depends on the FilOzone SDK stack, and `@filoz/synapse-core/abis` exposes `pdp` and `fwss` ABI exports. Since `graph codegen` needs JSON files on disk, we could add a small pre-codegen script that writes `apps/subgraph/abis/PDPVerifier.json` and `apps/subgraph/abis/FilecoinWarmStorageService.json` from the package before `graph codegen`. That would avoid ABI drift and make future contract updates a dependency bump instead of a large copied diff.

SgtPooki (changes requested):
one minor comment that might be based on current subgraph failures.. and another that should hopefully reduce the size of changes needed significantly (use source ABIs)

apps/subgraph/package.json
dennis-tra: done in: https://github.com/FilOzone/dealbot/pull/469/changes/5835eef31fc0056b71804287dc34b3913dcb29b3

apps/subgraph/src/pdp-verifier.ts
dennis-tra: Oh this logic was copied from the pdp-explorer:

https://github.com/FilOzone/pdp-explorer/blob/2bd606f087efe43998e212287e779e542863db33/subgraph/src/pdp-verifier.ts#L837

This means the current deployment is also affected by this? I guess whatever fix will land here should eventually also be ported to the pdp-explorer?

dennis-tra: done in https://github.com/FilOzone/dealbot/pull/469/changes/fc95fdf0913d63d43e293b432bb2e73ec9d3a72d

SgtPooki: @silent-cipher can you confirm this is an actual bug or am i misunderstanding things?

SgtPooki (approved):

apps/subgraph/networks.json
BigLep, line 1: I agree these values shouldn't change, but can we either import them from https://github.com/FilOzone/filecoin-services/blob/main/service_contracts/deployments.json or at least fail the build if they differ so that humans know to inspect?

apps/subgraph/README.md
BigLep, line 1: A couple of things:

1. I think this should explain why we are maintaing our own subgraph vs. using filezone/pdp-explorer and https://github.com/FIL-Builders/fwss-subgraph.

2. Lets also document which checks depend on which subgraph(s).

.github/workflows/test.yml
```diff
@@ -79,6 +79,9 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
+      - name: Codegen subgraph
+        run: pnpm --filter @dealbot/subgraph codegen
```
BigLep, line 83: I'm being lazy here just to make sure I get this review out before I sign off, but is there any assertion here with this? Under what circumstances will this fail?

Baiscally I would want to make sure we catch in CI
- codegen completes successfully
- our typescript code compiles / works with the generated code (no compiler errors)
- any tests we've written for the subragph pass

Is all that happening?

apps/subgraph/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+# Full build for one network
+pnpm build:mainnet
+pnpm build:calibration
+
+# Run matchstick tests
+pnpm test
+```
+
+## Deploy
```
BigLep, line 34: Dealbot has some extensive release process notes (https://github.com/FilOzone/dealbot/blob/main/docs/release-process.md ), and I think we need to have more cross linking between them.

I think we need to provide more details around sequencing...

In the case where the subgraph has a backwards compatible/additive change, I assume the sequnce of events is:
1. Deploy subgraph
2. Update dealbot to use new subgraph release
3. Deploy dealbot
(we should spell out how this works with staging and production)

I guess the set of rules apply for if we have a backwards incompatible change. A new sugraph version has to get deployed first and then dealbot can be  updated and deployed.

Anyways, I'd like to make sure we're thinkinga bout the subgraph as an individual component, and I assume it needs to be versioned independent of the dealbot backend, api, frontend, etc.

There is probably inspiration to draw from from filecoin-pay-explorer: https://github.com/FilOzone/filecoin-pay-explorer/blob/main/.github/ISSUE_TEMPLATE/release.md . It would be great if we had similar paradigms for our our subgraph deployments.  For example, I think it would be great if/hhen a new subgraph version is released that it automatically triggers a new subgraph deployment (like we have with filecoin-pay-explorer).

filecoin-pay-explorer uses tags (e.g., prod). If we don't think that's useful for us (I'm not sure, I need to think more on this), then we should at least document why we're treating dealbot's subgraph differently.

BigLep (commented):
I know I'm late to the party. I'm only engaging now as part of getting my head around  https://github.com/FilOzone/dealbot/pull/487 with the new subgraph.

I assume feedback here will get discussed here if needed, but ultimately incrporated into #487 or a separate PR.

.github/workflows/test.yml
dennis-tra: All of this is happening in this `test.yml` here.

This particular line does `codegen completes successfully` and the second and third bullet points are covered with `pnpm test` in L86 which also runs the subgraph tests.

apps/subgraph/networks.json
dennis-tra: Understood! I'm leaning towards an assertion in CI because while the contract addresses can be imported, the start blocks cannot. This means if CI breaks it requires manual intervention to also fix the start block values.

Secondly, we're already importing the synapse-sdk which also defines the contract addresses and we are already using this for the ABIs' source of truth anyway.

dennis-tra: Done in: https://github.com/FilOzone/dealbot/pull/542/changes/2875e28c7792b09745054e0ed66bcbebbb553d22

apps/subgraph/README.md
dennis-tra: Done in: https://github.com/FilOzone/dealbot/pull/542/changes/4ed7bc2b76c4189cc89e2c9ba0b5d0057413f8d5

apps/subgraph/README.md
dennis-tra: Attempted in https://github.com/FilOzone/dealbot/pull/542/changes/cb5d5ea9cb6509d96b4475526b9483b9c9a81a21

PR #504, opened 2026-05-04
feat: add pull flow testing

### LINKED ISSUE #300: Add pull flow testing to validate SP piece retrieval from external sources
DealBot should test the new SP "pull" pathway: the flow where an SP fetches a piece from an external URL rather than receiving it via direct upload. This is a critical path for multi-copy durability (where a secondary SP pulls from the primary or another source) and needs validation alongside existing data storage checks.

## Background

The current data storage check tests the direct upload path: DealBot uploads a piece to an SP, then verifies onchain confirmation, discoverability, and retrieval. But the SDK's multi-copy durability flow uses a different Curio pathway where the SP is told to pull a piece from a URL. This path has distinct failure modes and performance characteristics that aren't currently covered.

## Curio Pull Implementation

Curio exposes `POST /pdp/piece/pull` (see [curio/pdp/handlers_pull.go](https://github.com/filecoin-project/curio/blob/pdpv0/pdp/handlers_pull.go)). Key details:

Request format:

```json
{
  "extraData": "0x...",
  "dataSetId": 0,
  "recordKeeper": "0x...",
  "pieces": [
    {
      "pieceCid": "baga...",
      "sourceUrl": "https://example.com/api/piece/baga..."
    }
  ]
}
```

Synapse should have utilities for this (soon):
* https://github.com/FilOzone/synapse-sdk/pull/544: the synapse-core piece to interact with this API
* https://github.com/FilOzone/synapse-sdk/pull/593: the rest of the picture which also adds `pull()` to `StorageContext`, wrapping it all up

There is some additional source URL validation ([curio/pdp/pull_types.go](https://github.com/filecoin-project/curio/blob/pdpv0/pdp/pull_types.go)):
- Path must _end_ in `/piece/{pieceCid}`
- CID in path must match the `pieceCid` field in the request

The API returns per-piece status with a progression of: pending => inProgress => retrying => complete => failed

You can poll the API by repeating the same `POST /pdp/piece/pull` request (it's idempotent via `(service, sha256(extraData), dataSetId, recordKeeper)` key). The SP returns current status for each piece. So you can hit it repeatedly and wait for success. This is built in to synapse-core with a "waitFor" function.

Pull task mechanics ([curio/tasks/pdp/task_pull_piece.go](https://github.com/filecoin-project/curio/blob/pdpv0/tasks/pdp/task_pull_piece.go)):
- Background task polls every 10s for pending items
- Downloads piece from source URL
- Computes CommP and verifies against expected CID
- Stores in StashStore, creates parked_pieces entry, StorePiece task then moves to "long-term" storage
- Max 5 retries with exponential backoff (10s base, 2x factor, 5min cap)
- 1 hour download timeout

## Discrete Testable Units

The pull flow decomposes into three independently testable operations:

1. SP can receive + park an uploaded piece: already tested by data storage check
2. SP can pull from a URL + park: new, this issue
3. SP can do add-pieces on chain for a parked piece: already tested by data storage check

Unit 2 is the gap. Once a piece is parked via pull, the add-pieces flow is identical to direct upload.

## Possible Approach: DealBot-Hosted Piece Endpoint

(From chat with @SgtPooki) DealBot hosts a temporary piece retrieval endpoint on the existing backend.

- Serve via `/api/piece/{pieceCid}`. This routes through Caddy to the Node backend already (the Caddy container can't dynamically serve new assets, but `/api/*`
forwards to the backend)
- Generate the random piece, convert to CAR, compute pieceCID
- Enable the endpoint for a specific pieceCID, limited time
- Tell the SP to pull from https://dealbot.filoz.org/api/piece/{pieceCid}
- Poll the SP for status until complete or failed
- Clean up the endpoint + stored piece data after confirmation

Lastly we want to verify the piece is actually there: this is tricky because the SP can claim it has the piece but do we believe it? We have two options: either download the piece back from the SP and check that the bytes are what we want (hash or just byte compare to original), or just by proceeding to AddPieces and make the SP prove it we get a level of assurance that's probably acceptable.

An alternative flow exists where we could use an existing SP and do a proper SP-to-SP pull. We don't even need to do an AddPieces, we just use a second SP as a staging ground and expect it to GC the piece. But this has a few problems:
- Ambiguity: if pull fails, the SP is the problem, not the source
- No dependency on another SP being healthy
- Tighter metrics: DealBot _could_ measure time-to-first-byte fetched and throughput from its own side

## Metrics

We coudl add new metrics for pull checks, some ideas:

* `pullRequestMs`: DealBot => SP pull request latency (initial POST
* `pullCompletionMs`: Pull request => SP reports complete
* `pullFirstByteMs`: Time from SP connecting to DealBot endpoint to first byte served (if DealBot-hosted)
* `pullThroughputBps`: Bytes served / time (if DealBot-hosted)
* `pullStatus`: Counter with value label: `complete`, `failed`, `retrying` (would need deeper control than the higher level Synapse APIs)

## Tasks
- [ ] Create data-storage-via-pull-check
- [ ] Enhance retrieval check with `/piece` retrieavl
- [ ] Create/update docs in https://github.com/FilOzone/dealbot/blob/main/docs/checks

silent-cipher:
closes #300


apps/backend/src/jobs/jobs.service.spec.ts
```diff
@@ -95,18 +96,18 @@ describe("JobsService schedule rows", () => {
...
+      oldestQueuedAgeGauge: { set: vi.fn() } as unknown as JobsServiceDeps[11],
+      oldestInFlightAgeGauge: { set: vi.fn() } as unknown as JobsServiceDeps[12],
+      jobsInFlightGauge: { set: vi.fn() } as unknown as JobsServiceDeps[13],
+      jobsEnqueueAttemptsCounter: { inc: vi.fn() } as unknown as JobsServiceDeps[14],
+      jobsStartedCounter: { inc: vi.fn() } as unknown as JobsServiceDeps[15],
+      jobsCompletedCounter: { inc: vi.fn() } as unknown as JobsServiceDeps[16],
+      jobsPausedGauge: { set: vi.fn() } as unknown as JobsServiceDeps[17],
+      jobDuration: { observe: vi.fn() } as unknown as JobsServiceDeps[18],
+      storageProvidersActive: { set: vi.fn() } as unknown as JobsServiceDeps[19],
+      storageProvidersTested: { set: vi.fn() } as unknown as JobsServiceDeps[20],
```
SgtPooki, line 110: us having to update this every time we add something new is quite annoying.

docs/jobs.md
SgtPooki: we should probably update the calculations in here so we have a clear understanding of limits given our defaults.

docs/environment-variables.md
```diff
@@ -140,6 +141,29 @@ DEALBOT_HOST=0.0.0.0
...
+- **Type**: `string` (URL)
+- **Required**: No (but required for [Pull Check](./checks/pull-check.md) when SPs cannot reach `DEALBOT_HOST:DEALBOT_PORT` directly)
+- **Default**: Empty (falls back to `http://${DEALBOT_HOST}:${DEALBOT_PORT}`)
+
+**Role**: Public base URL for the Dealbot HTTP API. Used to construct the hosted-piece source URL handed to a storage provider during a pull check (`{DEALBOT_API_PUBLIC_URL}/api/piece/{pieceCid}`).
+
+**When to update**:
+
+- Set to the public URL of your Dealbot deployment whenever pull checks are enabled and SPs cannot reach the bind address directly (the typical production case)
+- Leave unset for local development where SPs reach Dealbot on `localhost`
```
SgtPooki, line 155: > - Leave unset for local development where SPs reach Dealbot on `localhost`

for local development, SPs can't reach us on localhost. Do we have a method for testing the pull_check locally that will work?

docs/environment-variables.md
```diff
@@ -907,6 +931,98 @@ Only used when `DEALBOT_JOBS_MODE=pgboss`.
...
+
+### `PULL_CHECK_JOB_TIMEOUT_SECONDS`
+
+- **Type**: `number` (seconds)
+- **Required**: No
+- **Default**: `360` (6 minutes)
+- **Minimum**: `60`
+- **Enforced**: Yes (config validation)
+
+**Role**: Maximum runtime for a pull-check job before forced abort via `AbortController`. Bounds the polling window for terminal SP pull status, the on-chain commit, and the direct `/piece/{pieceCid}` re-fetch combined.
```
SgtPooki: why are we exercising on-chain commits for the pull check? I don't think we need to enforce or even check that.. from issue #300:

> Discrete Testable Units
> The pull flow decomposes into three independently testable operations:
> 
> > SP can receive + park an uploaded piece: already tested by data storage check
> > SP can pull from a URL + park: new, this issue
> > SP can do add-pieces on chain for a parked piece: already tested by data storage check


> > Unit 2 is the gap. Once a piece is parked via pull, the add-pieces flow is identical to direct upload.

we should be targeting just the "SP can pull from a URL + park: new, this issue" discrete unit

SgtPooki (commented):
I haven't looked through it all, but i'm mainly concerned about the mention of on-chain check which is why i'm submitting this review early.

the concern i have is the flow implemented from #300 w.r.t. this step:

> verify the piece is actually on SP: this is tricky because the SP can claim it has the piece but do we believe it? We have two options: either download the piece back from the SP and check that the bytes are what we want (hash or just byte compare to original), or just by proceeding to AddPieces and make the SP prove it we get a level of assurance that's probably acceptable

did you look into doing "download the piece back from the SP and check that the bytes are what we want"?

or is the "addpieces" simpler?

docs/environment-variables.md
silent-cipher: Yeah, SPs can’t reach us on localhost, so I tested it on a vps with dealbot exposed via an nginx proxy.

docs/environment-variables.md
silent-cipher: yeah, makes sense. We don’t get anything valuable from the commit, so I’ll skip it and directly download the piece for validation.

silent-cipher:
> did you look into doing "download the piece back from the SP and check that the bytes are what we want"?

Actually, after the commit, I download the piece back from the SP, recompute the piece cid, and verify that it matches the original piece cid uploaded via dealbot. So I’ll skip the commit for now.

docs/jobs.md
silent-cipher: [4267cae](https://github.com/FilOzone/dealbot/pull/504/commits/4267cae2d6b48d1678ac76dd287a2911a2a67e77)

docs/environment-variables.md
silent-cipher: Removed commit in [5f4e729](https://github.com/FilOzone/dealbot/pull/504/commits/5f4e729f9608f19e4c83896afd31472a4d81c977)

apps/backend/src/jobs/jobs.service.ts
```diff
@@ -645,6 +657,72 @@ export class JobsService implements OnModuleInit, OnApplicationShutdown {
...
+        "Pull check job skipped: provider is blocked for scheduled pull checks",
+      );
+      if (logContext == null) {
+        clearTimeout(timeoutId);
+        return "success";
+      }
+      try {
+        await this.pullCheckService.runPullCheck(spAddress, abortController.signal, logContext);
+        return "success";
+      } catch (error) {
```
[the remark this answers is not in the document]
silent-cipher: I can back HostedPieceRegistry with Postgres so any replica can resolve hosted piece metadata.

The part I’m less sure about is sharing the actual piece bytes across instances, since they’re currently stored locally on the worker filesystem.

@SgtPooki, could you help suggest a good approach for that?

SgtPooki:
> > did you look into doing "download the piece back from the SP and check that the bytes are what we want"?
> 
> 
> 
> Actually, after the commit, I download the piece back from the SP, recompute the piece cid, and verify that it matches the original piece cid uploaded via dealbot. So I’ll skip the commit for now.

And now the part about piece cleanup isnt a concern any longer, correct?

silent-cipher:
> And now the part about piece cleanup isnt a concern any longer, correct?

Yeah, since we no longer commit the piece, cleanup is no longer needed.

BigLep:
2026-05-07 standup discussion notes: https://github.com/FilOzone/dealbot/issues/300#issuecomment-4398784304

apps/backend/src/pull-check/pull-check.service.ts
```diff
@@ -0,0 +1,287 @@
...
+        pieceSizeBytes: prepared.registration.size,
+      });
+    } catch (error) {
+      this.pullCheckMetrics.recordStatus(labels, classifyFailureStatus(error));
+      throw error;
+    } finally {
+      if (prepared) {
+        await this.pullPieceRepository.forget(prepared.registration.pieceCid);
+      }
+    }
```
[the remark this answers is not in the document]
silent-cipher: There doesn't seem to be any reason for keeping entry in DB once job completes processing.

docs/checks/pull-check.md
```diff
@@ -0,0 +1,155 @@
...
+  Poll -->|other terminal status| Fail["Mark pull check failed"]
+  Validate -->|matches| Success["Mark pull check successful"]
+  Validate -->|mismatch or fetch error| Fail
+  Success --> Cleanup
+  Fail --> Cleanup["Forget hosted piece registration"]
+```
+
+### 1. Prepare the hosted piece
+
+Dealbot computes a deterministic PieceCID for a synthetic test piece and registers it in the Postgres `pull_pieces` table. The registration carries a TTL controlled by `PULL_CHECK_HOSTED_PIECE_TTL_SECONDS` so the source remains available for the entire pull window.
```
SgtPooki: `PULL_CHECK_HOSTED_PIECE_TTL_SECONDS` var is not implemented? only mention is in this doc in two places.

docs/checks/pull-check.md
```diff
@@ -0,0 +1,155 @@
...
+
+Source: [`pull-check.service.ts`](../../apps/backend/src/pull-check/pull-check.service.ts)
+
+## Pull Check Status Progression
+
+A pull check has a single terminal status, recorded once per check via [`pullCheckStatus`](./events-and-metrics.md#pullCheckStatus):
+
+| Overall Status | Meaning |
+|--------|---------|
+| `success` | All five [assertions](#what-gets-asserted) passed within the job timeout. |
```
SgtPooki: there are only 4 assertions. remove count 
```suggestion
| `success` | All [assertions](#what-gets-asserted) passed within the job timeout. |
```

docs/checks/pull-check.md
```diff
@@ -0,0 +1,155 @@
...
+
+In addition to the overall status, dealbot records the **raw SP-reported terminal pull status** via [`pullCheckProviderStatus`](./events-and-metrics.md#pullCheckProviderStatus) (for example `complete`, `failed`, `not_found`). This separates "SP said it failed" from "dealbot's downstream validation failed" in dashboards.
+
+## HTTP API
+
+The dealbot API exposes one endpoint dedicated to pull checks:
+
+| Method | Path | Description |
+|--------|------|-------------|
+| `GET` | `/api/piece/{pieceCid}` | Streams the temporary hosted piece bytes for an in-flight pull check. Returns `200` with the bytes when an active registration exists, `410 Gone` when the registration has been cleaned up or expired, and `404 Not Found` when no registration exists. |
```
SgtPooki: Code only has 404 currently?

docs/checks/pull-check.md
```diff
@@ -0,0 +1,155 @@
...
+
+The endpoint is registered on the same `/api` prefix as the other dealbot HTTP endpoints. It is intentionally unauthenticated because SPs must be able to pull from it during a check; access is bounded by the per-piece TTL.
```
SgtPooki: missing TTL env var

docs/checks/pull-check.md
```diff
@@ -0,0 +1,155 @@
...
+## Configuration
+
+Key environment variables that control pull check behavior:
+
+| Variable | Description |
+|----------|-------------|
+| `DEALBOT_API_PUBLIC_URL` | Public base URL used to construct the hosted-piece source URL handed to SPs. Required for any deployment where SPs cannot reach `DEALBOT_HOST:DEALBOT_PORT` directly. |
+| `PULL_CHECKS_PER_SP_PER_HOUR` | Per-SP pull check rate. |
+| `PULL_CHECK_JOB_TIMEOUT_SECONDS` | Max end-to-end pull check job runtime before forced abort. |
+| `PULL_CHECK_HOSTED_PIECE_TTL_SECONDS` | TTL of the temporary hosted piece source served at `/api/piece/{pieceCid}`. |
```
SgtPooki: non-existent in code?

docs/environment-variables.md
```diff
@@ -907,6 +931,82 @@ Only used when `DEALBOT_JOBS_MODE=pgboss`.
...
+- **Minimum**: `1024`
+
+**Role**: Size of the synthetic random piece dealbot generates per pull check. The same byte length is used to compute [`pullCheckThroughputBps`](./checks/events-and-metrics.md#pullCheckThroughputBps).
+
+**When to update**:
+
+- Decrease for quicker, lower-bandwidth pull tests
+- Increase to stress-test the SP's outbound fetch throughput
+
+**Note**: Pull-check pieces are committed on-chain but **not** tracked in the `deals` table, so they are not garbage-collected by [Piece Cleanup](#piece-cleanup). Larger pieces accrue on the SP unless removed manually.
```
SgtPooki: pieces are no longer committed on chain

apps/backend/src/pull-check/pull-piece.controller.ts
```diff
@@ -0,0 +1,71 @@
...
+  private readonly logger = new Logger(PieceSourceController.name);
+
+  constructor(
+    private readonly pullCheckService: PullCheckService,
+    private readonly pullPieceRepository: PullPieceRepository,
+  ) {}
+
+  @Get("piece/:pieceCid")
+  @ApiResponse({ status: 200, description: "Raw piece bytes streamed to the caller" })
+  @ApiResponse({ status: 404, description: "No active pull piece exists for this pieceCid" })
```
SgtPooki, line 31: either drop 410 claims from doc or implement

apps/backend/src/pull-check/pull-piece.controller.ts
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Serves the temporary pull-piece bytes that a storage provider must fetch
+ * during a pull check. Bound to the same `/api/*` prefix as other DealBot HTTP
+ * endpoints. The path component must end with `/piece/{pieceCid}` so that
+ * SP-side pull workers can address the resource directly.
+ */
+@ApiTags("Pull Check")
+@Controller("api")
+export class PieceSourceController {
+  private readonly logger = new Logger(PieceSourceController.name);
```
SgtPooki, line 20: we likely need some sane rate-limits or throttler guard to prevent nefarious actors from trying to take down dealbot by spamming this endpoint. 

we could use `@nestjs/throttler` here (i'm not familiar with it) to per-IP rate limit..

we could also prevent concurrent streams to this endpoint based on # sps and pull_checks_per_sp_per_hour.. probably overkill, but is worth thinking about.

apps/backend/src/pull-check/pull-piece.controller.ts
```diff
@@ -0,0 +1,71 @@
...
+      });
+      res.status(404).send("Pull piece source not found");
+      return;
+    }
+
+    const { registration, stream } = opened;
+    res.setHeader("Content-Type", "application/octet-stream");
+    res.setHeader("Content-Length", registration.size.toString());
+    res.setHeader("Cache-Control", "no-store");
+    res.setHeader("X-Pull-Check-Piece-CID", registration.pieceCid);
```
SgtPooki: do we need to support accept-ranges? If not, set "Accept-Ranges: none" as well

apps/backend/src/pull-check/pull-check.service.ts
```diff
@@ -0,0 +1,284 @@
...
+        throughputBps,
+        pieceSizeBytes: prepared.registration.size,
+      });
+    } catch (error) {
+      this.pullCheckMetrics.recordStatus(labels, classifyFailureStatus(error));
+      throw error;
+    } finally {
+      if (prepared) {
+        await this.pullPieceRepository.forget(prepared.registration.pieceCid);
+      }
```
SgtPooki, line 188: are we waiting until we "forget" the piece? relevant to missing TTL env var

SgtPooki (changes requested):

docs/checks/pull-check.md
```diff
@@ -0,0 +1,155 @@
...
+
+Source code links throughout this document point to the current implementation.
+
+For event and metric definitions used by the dashboard, see [Dealbot Events & Metrics](./events-and-metrics.md).
+
+## Overview
+
+A "pull check" exercises the **storage provider pull-to-park pathway**: dealbot publishes a temporary piece at `/api/piece/{pieceCid}`, asks the SP to fetch (pull) and park it via the Synapse `pullPieces` API, waits for a terminal SP pull status, and finally re-fetches the piece from the SP to verify byte-for-byte integrity.
+
+The pull check answers a different question than the [Data Storage check](./data-storage.md): instead of *uploading* bytes to the SP, it asks the SP to *pull* bytes from a public URL. This validates an SP's outbound HTTP fetcher, the pull request lifecycle, and retrieval surface.
```
BigLep:
```suggestion
The pull check answers a different question than the [Data Storage check](./data-storage.md): instead of *uploading* bytes to the SP, it asks the SP to *pull* bytes from a public URL. This validates an SP's outbound HTTP fetcher, the pull request lifecycle, and `/piece` retrieval surface. (`/piece` retrieval is not covered by the [Data Storage check](./data-storage.md).)
```

docs/checks/pull-check.md
```diff
@@ -0,0 +1,155 @@
...
+
+A successful pull check requires all [assertions in the table below](#what-gets-asserted) to pass. Failure occurs if any step fails or the job exceeds its max allowed time. Operational timeouts exist to prevent jobs from running indefinitely, but they are not quality assertions.
+
+> **Where results live:** Pull check results are exported to Prometheus and structured logs only. They are **not** persisted in Postgres or written to ClickHouse.
```
BigLep, line 17: Ideally there wouldn't be any special callouts here. Any issue with us writing to clickhouse?

docs/checks/pull-check.md
```diff
@@ -0,0 +1,155 @@
...
+
+## What Gets Asserted
+
+Each pull check asserts the following for every SP:
+
+| # | Assertion | How It's Checked | Retries | Relevant Metric for Setting a Max Duration | Implemented? |
+|---|-----------|------------------|:---:|--------------------------------------------|:---:|
+| 1 | SP accepts the pull request | `pullPieces` returns without error and reports a non-terminal-failure status | 0 | [`pullCheckRequestLatencyMs`](./events-and-metrics.md#pullCheckRequestLatencyMs) | Yes |
```
BigLep:
```suggestion
| 1 | SP accepts the pull request | Synapse `pullPieces` (i.e., initial call to SP `POST /pdp/piece/pull`) returns without error and reports a non-terminal-failure status | 0 | [`pullCheckRequestLatencyMs`](./events-and-metrics.md#pullCheckRequestLatencyMs) | Yes |
```

docs/checks/pull-check.md
```diff
@@ -0,0 +1,155 @@
...
+| 2 | SP reaches a terminal `complete` pull status | `waitForPullStatus` polls the SP until a terminal status is reported | Polling with delay until [`PULL_CHECK_JOB_TIMEOUT_SECONDS`](../environment-variables.md#pull_check_job_timeout_seconds) | [`pullCheckCompletionLatencyMs`](./events-and-metrics.md#pullCheckCompletionLatencyMs) | Yes |
```
BigLep:
```suggestion
| 2 | SP reaches a terminal `complete` pull status | Synapse `waitForPullStatus` polls the SP (using `POST /pdp/piece/pull`) until a terminal status is reported | Polling with contune until [`PULL_CHECK_JOB_TIMEOUT_SECONDS`](../environment-variables.md#pull_check_job_timeout_seconds) | [`pullCheckCompletionLatencyMs`](./events-and-metrics.md#pullCheckCompletionLatencyMs) | Yes |
```

docs/checks/pull-check.md
```diff
@@ -0,0 +1,155 @@
...
+  Generate["Compute PieceCID + register hosted source in Postgres<br/>at /api/piece/{pieceCid}"]
+  Generate --> Submit["Submit pullPieces request to SP"]
+  Submit --> Poll["Poll SP via waitForPullStatus<br/>until terminal pull status"]
+  Poll -->|complete| Validate["Direct /piece/{pieceCid} fetch from SP<br/>+ recompute pieceCid"]
+  Poll -->|other terminal status| Fail["Mark pull check failed"]
+  Validate -->|matches| Success["Mark pull check successful"]
+  Validate -->|mismatch or fetch error| Fail
+  Success --> Cleanup
+  Fail --> Cleanup["Forget hosted piece registration"]
+```
```
BigLep, line 46:
```suggestion
```mermaid
flowchart TD
  Generate["Compute PieceCID + register hosted source in Postgres<br/>at /api/piece/{pieceCid}"]
  Generate --> Submit["Submit pullPieces request to SP"]
  Submit --> |SP responds with HTTP 200|Poll["Poll SP via waitForPullStatus<br/>until terminal pull status"]
  Submit --> |SP doesn't respond with HTTP 200| Fail["Mark pull check failed"]
  Poll -->|complete| Validate["Direct /piece/{pieceCid} fetch from SP<br/>+ recompute pieceCid"]
  Poll -->|other terminal status| Fail
  Validate -->|matches| Success["Mark pull check successful"]
  Validate -->|mismatch or fetch error| Fail
  Success --> Cleanup
  Fail --> Cleanup["Forget hosted piece registration"]
```
```

My understanding is that there is another failure that can occur if the initial pull request isn't accepted.

File: docs/checks/pull-check.md [hunk 21]
```diff
@@ -0,0 +1,155 @@
...
+
+### 1. Prepare the hosted piece
+
+Dealbot computes a deterministic PieceCID for a synthetic test piece and registers it in the Postgres `pull_pieces` table. The registration carries a TTL controlled by `PULL_CHECK_HOSTED_PIECE_TTL_SECONDS` so the source remains available for the entire pull window.
+
+By persisting registrations to Postgres instead of in-memory, the hosted source can be resolved by any API pod in a multi-pod deployment, even if the pull check was initiated by a different worker pod.
+
+The synthetic data is **not** stored on disk. Instead, dealbot uses a deterministic pseudo-random generator (AES-256-CTR) to stream the same bytes whenever the SP fetches the piece or dealbot needs to re-compute the CID for validation.
```
BigLep:
```suggestion
Dealbot computes a deterministic PieceCID for a synthetic test piece.  The synthetic data is **not** stored on disk. Instead, dealbot uses a deterministic pseudo-random generator (AES-256-CTR) to generate it as needed. 

The pieceCid and synthetic piece's random seed are registered it in the Postgres `pull_pieces` table. The registration carries a TTL controlled by `PULL_CHECK_HOSTED_PIECE_TTL_SECONDS` so the source remains available for the entire pull window.

By persisting registrations to Postgres instead of in-memory, the hosted source can be resolved by any API pod in a multi-pod deployment, even if the pull check was initiated by a different worker pod. It generates the same piece and streams the whenever the SP fetches the piece or dealbot needs to re-compute the CID for validation.
```

BigLep: I think we should we should talk about how the pieceCid is constructed before we talk about storing it to the DB.

docs/checks/pull-check.md
```diff
@@ -0,0 +1,155 @@
...
+
+Dealbot calls `pullPieces` from `@filoz/synapse-core/sp` with the pieceCid, the source URL, and either the SP's existing `dataSetId`/`clientDataSetId` or the SP `payee` for new-dataset flows. The submission timestamp is stamped on the registration so it can later be subtracted from the first-byte event.
+
+Source: [`pull-check.service.ts` (`runPullCheck`)](../../apps/backend/src/pull-check/pull-check.service.ts)
+
+### 3. Wait for terminal SP pull status
+
+`waitForPullStatus` polls the SP at `PULL_CHECK_POLL_INTERVAL_SECONDS` until the SP reports a terminal status (`complete` or `failed`) or the job timeout fires. Dealbot increments the [`pullCheckProviderStatus`](./events-and-metrics.md#pullCheckProviderStatus) counter exactly once with the **terminal** status; intermediate poll statuses are not counted.
+
+When the SP fetches `/api/piece/{pieceCid}` for the first time, the controller stamps a first-byte timestamp on the registration. This is the basis for [`pullCheckFirstByteMs`](./events-and-metrics.md#pullCheckFirstByteMs).
```
BigLep:
```suggestion
When dealbot receives the SP request for `/api/piece/{pieceCid}` for the first time, dealbot stamps a first-byte timestamp on the registration. This is the basis for [`pullCheckFirstByteMs`](./events-and-metrics.md#pullCheckFirstByteMs).

In parallel, dealbot`waitForPullStatus` polls the SP at `PULL_CHECK_POLL_INTERVAL_SECONDS` until the SP reports a terminal status (`complete` or `failed`) or the job timeout fires. Dealbot increments the [`pullCheckProviderStatus`](./events-and-metrics.md#pullCheckProviderStatus) counter exactly once with the **terminal** status; intermediate poll statuses are not counted.


```

I removed the "controller" logic, unless that is covered elsewhere.

I rearranged the paragraphs to go in chronological order of the events.

docs/checks/pull-check.md
```diff
@@ -0,0 +1,155 @@
...
+
+Source: [`pull-piece.controller.ts`](../../apps/backend/src/pull-check/pull-piece.controller.ts)
+
+### 4. Direct piece-fetch validation
+
+After commit, dealbot fetches `{serviceURL}/piece/{pieceCid}` from the SP, re-computes the piece CID over the response body, and compares it against the expected CID. A mismatch fails the pull check with `failure.other`. A network or HTTP error during validation also fails the check (transport errors are intentionally not retried).
+
+Aborts (job timeout) propagate as throws and are classified as `failure.timedout` rather than as a validation failure.
```
BigLep:
```suggestion
After `waitForPullStatus`, dealbot fetches `{serviceURL}/piece/{pieceCid}` from the SP, re-computes the piece CID over the response body, and compares it against the expected CID. A mismatch fails the pull check with `failure.other`. A network or HTTP error during validation also fails the check (transport errors are intentionally not retried).

Aborts (job timeout) propagate as throws and are classified as `failure.timedout` rather than as a validation failure.
```

I wasn't sure what was meant by "commit".

docs/checks/pull-check.md
```diff
@@ -0,0 +1,155 @@
...
+
+Source: [`pull-check.service.ts` (`validateByDirectPieceFetch`)](../../apps/backend/src/pull-check/pull-check.service.ts)
+
+### 5. Cleanup
+
+Whether the pull check succeeds or fails, the `finally` block removes the registration entry.
```
BigLep:
```suggestion
Whether the pull check succeeds or fails, the `finally` block removes the registration entry from postgres' database.
```

docs/checks/pull-check.md
```diff
@@ -0,0 +1,155 @@
...
+
+## FAQ
+
+### Why isn't the pull-check piece tracked in the `deals` table?
+
+Pull checks are intentionally isolated from the data-storage flow: they don't pass through `DealService.createDeal`, don't allocate a `Deal` entity. This keeps the pull-check signal independent of the data-storage success rate.
+
+### Why does a "cached pull" not record `pullCheckFirstByteMs`?
+
+If an SP previously pulled the same piece CID and serves the new pull request from a local cache, it will never fetch `/api/piece/{pieceCid}`, so dealbot has no first-byte timestamp to subtract. In that case dealbot skips the histogram observation rather than emit a misleading zero. Cached pulls are uncommon today because each pull check generates a fresh random piece, but the registry's first-byte capture is **idempotent** so retried pulls during a single check do not skew measurements either.
```
BigLep: What would even cause this to happen? Don't we generate a new CID for each pull check? If this an edge case because we happen to generate the same seed, then I think we can skip this here and in code. That's an extreme edge case (assuming our seed space is large).

docs/checks/pull-check.md
```diff
@@ -0,0 +1,155 @@
...
+## Metrics Recorded
+
+Metric definitions (including Prometheus metrics) live in [Dealbot Events & Metrics](./events-and-metrics.md). The metrics emitted by a pull check are:
+
+- [`pullCheckRequestLatencyMs`](./events-and-metrics.md#pullCheckRequestLatencyMs)
+- [`pullCheckCompletionLatencyMs`](./events-and-metrics.md#pullCheckCompletionLatencyMs)
+- [`pullCheckFirstByteMs`](./events-and-metrics.md#pullCheckFirstByteMs) (only when the SP actually pulled from `/api/piece/{pieceCid}`)
+- [`pullCheckThroughputBps`](./events-and-metrics.md#pullCheckThroughputBps)
+- [`pullCheckStatus`](./events-and-metrics.md#pullCheckStatus)
+- [`pullCheckProviderStatus`](./events-and-metrics.md#pullCheckProviderStatus)
```
BigLep:
```suggestion
- [`pullCheckRequestLatencyMs`](./events-and-metrics.md#pullCheckRequestLatencyMs)
- [`pullCheckFirstByteMs`](./events-and-metrics.md#pullCheckFirstByteMs) (only when the SP actually pulled from `/api/piece/{pieceCid}`)
- [`pullCheckCompletionLatencyMs`](./events-and-metrics.md#pullCheckCompletionLatencyMs)
- [`pullCheckThroughputBps`](./events-and-metrics.md#pullCheckThroughputBps)
- [`pullCheckStatus`](./events-and-metrics.md#pullCheckStatus)
- [`pullCheckProviderStatus`](./events-and-metrics.md#pullCheckProviderStatus)
```

I like to discuss things in chronologoical order.

And per below, what is the scenario where an SP won't call `/api/piece/{pieceCid}`?

docs/checks/pull-check.md
```diff
@@ -0,0 +1,155 @@
...
+
+- **Data format:** Deterministic pseudo-random bytes
+- **Default size:** `PULL_CHECK_PIECE_SIZE_BYTES` (default 10 MiB)
+- **Source URL:** `{apiPublicUrl}/api/piece/{pieceCid}`
+
+Source: [`pull-check.service.ts` (`preparePullPiece`)](../../apps/backend/src/pull-check/pull-check.service.ts), [`pull-piece.repository.ts`](../../apps/backend/src/pull-check/pull-piece.repository.ts)
+
+### 2. Submit the pull request
+
+Dealbot calls `pullPieces` from `@filoz/synapse-core/sp` with the pieceCid, the source URL, and either the SP's existing `dataSetId`/`clientDataSetId` or the SP `payee` for new-dataset flows. The submission timestamp is stamped on the registration so it can later be subtracted from the first-byte event.
```
BigLep: what is " the SP's existing `dataSetId`/`clientDataSetId` or the SP `payee` for new-dataset flows"? This is the first I am seeing this discussed in this doc. This needs expansion.

docs/checks/events-and-metrics.md
```diff
@@ -55,13 +55,47 @@ sequenceDiagram
...
+  Dealbot->>SP: directPieceFetchStarted (/piece/{cid})
+  SP-->>Dealbot: directPieceFetchCompleted
+  Dealbot-->>Dealbot: pullCheckIntegrityChecked
+```
+
+### Pull Check Event List
+
+| Event | Definition | Implemented | Source of truth |
+|------|------------|:------:|-----------------|
+| <a id="pullRequestSubmitted"></a>`pullRequestSubmitted` | Dealbot calls `pullPieces` against the SP for the registered piece CID. | Yes | [`pull-check.service.ts`](../../apps/backend/src/pull-check/pull-check.service.ts) |
```
BigLep:
```suggestion
| <a id="pullRequestSubmittedToSp"></a>`pullRequestSubmittedToSp` | Dealbot calls Synpase pullPiece (`POST /pdp/piece/pull`) against the SP for the registered piece CID. | Yes | [`pull-check.service.ts`](../../apps/backend/src/pull-check/pull-check.service.ts) |
```

docs/checks/events-and-metrics.md
```diff
@@ -55,13 +55,47 @@ sequenceDiagram
...
+| <a id="pullRequestAcknowledged"></a>`pullRequestAcknowledged` | SP returns from `pullPieces` (success or non-terminal-failure). | Yes | [`pull-check.service.ts`](../../apps/backend/src/pull-check/pull-check.service.ts) |
```
BigLep:
```suggestion
| <a id="pullRequestAcknowledgedBySp"></a>`pullRequestAcknowledgedBySp` | SP returns from `pullPieces` (success or non-terminal-failure). | Yes | [`pull-check.service.ts`](../../apps/backend/src/pull-check/pull-check.service.ts) |
```

docs/checks/events-and-metrics.md
```diff
@@ -55,13 +55,47 @@ sequenceDiagram
...
+| <a id="hostedPieceFirstByteRead"></a>`hostedPieceFirstByteRead` | SP reads the first byte of `/api/piece/{pieceCid}` from dealbot. Recorded once per registration. | Yes | [`pull-piece.controller.ts`](../../apps/backend/src/pull-check/pull-piece.controller.ts) |
```
BigLep:
```suggestion
| <a id="pullRequestStartedBySp"></a>`pullRequestStartedBySp` | Dealbot receives SP request for `/api/piece/{pieceCid}`. Recorded once per registration. | Yes | [`pull-piece.controller.ts`](../../apps/backend/src/pull-check/pull-piece.controller.ts) |
```

docs/checks/events-and-metrics.md
```diff
@@ -55,13 +55,47 @@ sequenceDiagram
...
+| <a id="pullTerminalStatusReported"></a>`pullTerminalStatusReported` | SP reports a terminal pull status (`complete`, `failed`, ...) via `waitForPullStatus`. Intermediate poll statuses are not counted. | Yes | [`pull-check.service.ts`](../../apps/backend/src/pull-check/pull-check.service.ts) |
```
BigLep:
```suggestion
| <a id="pullRequestIsTerminal"></a>`pullRequestIsTerminal` | Dealbot determines the the pull request is in terminal pull status (`complete`, `failed`, ...) via `waitForPullStatus` or the polling operation has timed out. Intermediate poll statuses are not counted. | Yes | [`pull-check.service.ts`](../../apps/backend/src/pull-check/pull-check.service.ts) |
```

docs/checks/events-and-metrics.md
```diff
@@ -55,13 +55,47 @@ sequenceDiagram
...
+| <a id="pullCheckIntegrityChecked"></a>`pullCheckIntegrityChecked` | Direct `/piece/{pieceCid}` fetch from the SP returns bytes whose recomputed pieceCid matches the expected CID. | Yes | [`pull-check.service.ts`](../../apps/backend/src/pull-check/pull-check.service.ts) |
```
BigLep:
```suggestion
| <a id="pullRequestIntegrityChecked"></a>`pullRequestIntegrityChecked` | Dealbot performs direct `/piece/{pieceCid}` retrieval from the SP and confirms the bytes match the pieceCid. | Yes | [`pull-check.service.ts`](../../apps/backend/src/pull-check/pull-check.service.ts) |
```

docs/checks/events-and-metrics.md
```diff
@@ -87,6 +121,10 @@ sequenceDiagram
 | <a id="dataStorageCheckMs"></a>`dataStorageCheckMs` | Data Storage | [`uploadToSpStart`](#uploadToSpStart) | [`ipfsRetrievalIntegrityChecked`](#ipfsRetrievalIntegrityChecked) | Duration of a Data Storage check | |
 | <a id="retrievalCheckMs"></a>`retrievalCheckMs` | Retrieval | Retrieval check start | [`ipfsRetrievalIntegrityChecked`](#ipfsRetrievalIntegrityChecked) | Duration of a Retrieval check | |
 | <a id="dataSetCreationMs"></a>`dataSetCreationMs` | Data-Set Creation | Data-set creation uploadToSpStart | Data-set creation pieceConfirmed | Duration of one data-set creation with confirmed piece (all using `createDataSetWithPiece`) | [`deal.service.ts`](../../apps/backend/src/deal/deal.service.ts) |
+| <a id="pullCheckRequestLatencyMs"></a>`pullCheckRequestLatencyMs` | Pull | [`pullRequestSubmitted`](#pullRequestSubmitted) | [`pullRequestAcknowledged`](#pullRequestAcknowledged) | Time from `pullPieces` submission to SP request acknowledgement. | [`pull-check.service.ts`](../../apps/backend/src/pull-check/pull-check.service.ts) |
```
BigLep:
```suggestion
| <a id="pullRequestAcknowledgementLatencyMs"></a>`pullRequestAcknowledgementLatencyMs` | Pull | [`pullRequestSubmitted`](#pullRequestSubmitted) | [`pullRequestAcknowledged`](#pullRequestAcknowledged) | Time from `pullPieces` submission to SP request acknowledgement. | [`pull-check.service.ts`](../../apps/backend/src/pull-check/pull-check.service.ts) |
```

docs/checks/events-and-metrics.md
```diff
@@ -87,6 +121,10 @@ sequenceDiagram
...
+| <a id="pullCheckCompletionLatencyMs"></a>`pullCheckCompletionLatencyMs` | Pull | [`pullRequestSubmitted`](#pullRequestSubmitted) | [`pullTerminalStatusReported`](#pullTerminalStatusReported) | Time from `pullPieces` submission to terminal SP pull status. Observed once on success and once on failure. | [`pull-check.service.ts`](../../apps/backend/src/pull-check/pull-check.service.ts) |
```
BigLep:
```suggestion
| <a id="pullRequestCompletionLatencyMs"></a>`pullRequestCompletionLatencyMs` | Pull | [`pullRequestSubmitted`](#pullRequestSubmitted) | [`pullTerminalStatusReported`](#pullTerminalStatusReported) | Time from `pullPieces` submission to terminal SP pull status. Observed once on success and once on failure. | [`pull-check.service.ts`](../../apps/backend/src/pull-check/pull-check.service.ts) |
```

docs/checks/events-and-metrics.md
```diff
@@ -87,6 +121,10 @@ sequenceDiagram
...
+| <a id="pullCheckFirstByteMs"></a>`pullCheckFirstByteMs` | Pull | [`pullRequestSubmitted`](#pullRequestSubmitted) | [`hostedPieceFirstByteRead`](#hostedPieceFirstByteRead) | Time from `pullPieces` submission to the SP reading the first byte of `/api/piece/{pieceCid}`. Skipped (no observation) when the SP serves the pull from a local cache and never fetches from dealbot. | [`pull-check.service.ts`](../../apps/backend/src/pull-check/pull-check.service.ts), [`pull-piece.controller.ts`](../../apps/backend/src/pull-check/pull-piece.controller.ts) |
```
BigLep:
```suggestion
| <a id="pullRequestStartedMs"></a>`pullRequestStartedMs` | Pull | [`pullRequestSubmitted`](#pullRequestSubmitted) | [`hostedPieceFirstByteRead`](#hostedPieceFirstByteRead) | Time from `pullPieces` submission to the SP reading the first byte of `/api/piece/{pieceCid}`. Skipped (no observation) when the SP serves the pull from a local cache and never fetches from dealbot. | [`pull-check.service.ts`](../../apps/backend/src/pull-check/pull-check.service.ts), [`pull-piece.controller.ts`](../../apps/backend/src/pull-check/pull-piece.controller.ts) |
```

We should move this up a line since it happens earlier chronologicially.

docs/checks/events-and-metrics.md
```diff
@@ -87,6 +121,10 @@ sequenceDiagram
...
+| <a id="pullCheckThroughputBps"></a>`pullCheckThroughputBps` | Pull | n/a | n/a | `(pieceSizeBytes / pullCheckCompletionLatencyMs) * 1000`. Upper-bound on actual transfer rate because `pullCheckCompletionLatencyMs` includes SP-side scheduling and dealbot's polling cadence. | [`pull-check.service.ts`](../../apps/backend/src/pull-check/pull-check.service.ts) |
```
BigLep:
```suggestion
| <a id="pullRequestThroughputBps"></a>`pullRequestThroughputBps` | Pull | n/a | n/a | `(pieceSizeBytes / pullCheckCompletionLatencyMs) * 1000`. Upper-bound on actual transfer rate because `pullCheckCompletionLatencyMs` includes SP-side scheduling and dealbot's polling cadence. | [`pull-check.service.ts`](../../apps/backend/src/pull-check/pull-check.service.ts) |
```

docs/checks/events-and-metrics.md
BigLep, line 1: I suggested some new events and metric names when looking holistically at the terms for the pull check flow. 

I leaned into the idea of there is a pull requst that is submitted, acknowledged, started, and finished.

Whatever event/metric names we decide will need to be propagated through the rest of the doc and in code.

BigLep (commented):
Thanks for jumping on this. I just looked at the docs portion.

Doing another one of these checks is reminding me of how various things shake out from agreeing on the English documentation. I don't think any big changes here, but I recognize it's tedious to do renames. Lets lock in on the docs side so you can just rename once (if there is agreement to rename).

---
https://github.com/FilOzone/dealbot/issues/300#issuecomment-3955076026

> we still want to make sure that it gets indexed, advertised to IPNI, and has standard IPFS retrieval

I see we aren't doing that. It isn't a dealbreaker for me, but I think we should sadd a FAQ entry to the pull-check.md

```markdwon
### Do pull checks verify IPNI indexing and IPFS retrieval?
No, not currently. We are assuming that if an SP does IPNI indexing and IPFS retrieval with data storage checks that it also happens when a piece is pulled. 
```

apps/backend/src/pull-check/pull-piece.controller.ts
silent-cipher: [93d6179](https://github.com/FilOzone/dealbot/pull/504/commits/93d617941251c526dad327e8feee92c8c970c2ca)

apps/backend/src/pull-check/pull-check.service.ts
silent-cipher: Fixed in [1a420f6](https://github.com/FilOzone/dealbot/pull/504/commits/1a420f6f90a7b1fe05cf656678aa1ca34ea9ea0a)

I've removed TTL env completely, since we remove the piece entry at the end of job lifecycle.

docs/checks/pull-check.md
silent-cipher: Yes, this is extreme edge case will remove it from here and in code.

docs/checks/pull-check.md
silent-cipher: If the initial pull piece POST request fails due to sp fault, then SP won't call `/api/piece/{pieceCid}`.

docs/checks/pull-check.md
silent-cipher: This section is a bit outdated. We no longer pass `dataSetId`/`clientDataSetId`. Instead, we pass the `payee`, which is required by the sdk for validation inside curio. curio then authenticates it via an `eth_call` to the pdp verifier contract.

docs/checks/events-and-metrics.md
silent-cipher: I've updated names to suggested names for events and metrics in [445a0d6](https://github.com/FilOzone/dealbot/pull/504/commits/445a0d6c63b5b215fbcb50d6f3c5615fa471bff9)

apps/backend/src/jobs/jobs.service.ts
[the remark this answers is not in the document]
silent-cipher: This has been resolved in [540f03d](https://github.com/FilOzone/dealbot/pull/504/commits/540f03de477c9cdab3095b7998ea639ab5574e88)

docs/checks/pull-check.md
silent-cipher: https://github.com/FilOzone/dealbot/pull/522

apps/backend/src/pull-check/pull-piece.repository.ts
```diff
@@ -0,0 +1,93 @@
...
+      {
+        pieceCid: registration.pieceCid,
+        providerAddress: registration.providerAddress,
+        key: registration.key,
+        size: registration.size,
+        pullSubmittedAt: null,
+        firstByteAt: null,
+      },
+      ["pieceCid"],
+    );
```
SgtPooki, line 34: its possible this could persist in the DB for a while if a pull job fails mid-flight. we might want an "expires_at" field and an explicit cleanup

apps/backend/src/pull-check/pull-check.service.ts
```diff
@@ -0,0 +1,292 @@
...
+    providerInfo: PDPProviderEx,
+    pieceCid: string,
+    logContext: ProviderJobContext,
+    signal?: AbortSignal,
+  ): Promise<boolean> {
+    signal?.throwIfAborted();
+    const pieceFetchUrl = this.constructPieceFetchUrl(providerInfo.pdp.serviceURL, pieceCid);
+    try {
+      const response = await this.httpClientService.requestWithMetrics<Buffer>(pieceFetchUrl, { signal });
+      const calculatedPieceCid = calculate(response.data);
```
SgtPooki: A large response here can exhaust the worker's memory. we could do a size check against `Content-Length`, and then ideally we would calculate the PieceCid from streamed bytes rather than loading fully into memory.

apps/backend/src/pull-check/pull-piece.controller.ts
```diff
@@ -0,0 +1,106 @@
...
+  private readonly logger = new Logger(PieceSourceController.name);
+
+  constructor(
+    private readonly pullCheckService: PullCheckService,
+    private readonly pullPieceRepository: PullPieceRepository,
+    private readonly streamTracker: PullPieceStreamTracker,
+  ) {}
+
+  @Get("piece/:pieceCid")
+  @UseGuards(PullPieceThrottlerGuard)
```
SgtPooki:
```suggestion
  @UseGuards(ThrottlerGuard)
```

We shouldn't need the custom guard at all

apps/backend/src/pull-check/pull-piece-throttler.guard.ts
```diff
@@ -0,0 +1,10 @@
+import { ExecutionContext, Injectable } from "@nestjs/common";
+import { ThrottlerGuard } from "@nestjs/throttler";
+
+@Injectable()
+export class PullPieceThrottlerGuard extends ThrottlerGuard {
+  protected async throwThrottlingException(context: ExecutionContext): Promise<void> {
+    const res = context.switchToHttp().getResponse();
+    res.status(429).setHeader("Retry-After", "60").send("Too many requests");
+  }
+}
```
SgtPooki: lets remove this file completely. we shouldn't need it

SgtPooki (changes requested):
only a few changes required before I think I can give a ship-it..

apps/backend/src/pull-check/pull-piece-throttler.guard.ts
silent-cipher: [02f7ab9](https://github.com/FilOzone/dealbot/pull/504/commits/02f7ab9cdbd5f7d76a04f876af2985542962d1bb)

apps/backend/src/pull-check/pull-check.service.ts
silent-cipher: [1ba5cd6](https://github.com/FilOzone/dealbot/pull/504/commits/1ba5cd63843d9b27df053fb9617a59c1d7e5d386)

apps/backend/src/pull-check/pull-piece.repository.ts
silent-cipher: https://github.com/FilOzone/dealbot/pull/523

SgtPooki (approved):
pushed up a few small fixes and merged main make sure #519 pulled cleanly

docs/checks/events-and-metrics.md
```diff
@@ -55,13 +55,47 @@ sequenceDiagram
...
+| <a id="pullRequestAcknowledgedBySp"></a>`pullRequestAcknowledgedBySp` | SP returns from `pullPieces` (success or non-terminal-failure). | Yes | [`pull-check.service.ts`](../../apps/backend/src/pull-check/pull-check.service.ts) |
+| <a id="pullRequestStartedBySp"></a>`pullRequestStartedBySp` | Dealbot receives SP request for `/api/piece/{pieceCid}` from dealbot. Recorded once per registration. | Yes | [`pull-piece.controller.ts`](../../apps/backend/src/pull-check/pull-piece.controller.ts) |
+| <a id="pullRequestIsTerminal"></a>`pullRequestIsTerminal` | Dealbot determines the pull request is in terminal pull status (`complete`, `failed`, ...) via `waitForPullPieces` or the polling operation has timed out. Intermediate poll statuses are not counted. | Yes | [`pull-check.service.ts`](../../apps/backend/src/pull-check/pull-check.service.ts) |
+| <a id="pullRequestIntegrityChecked"></a>`pullRequestIntegrityChecked` | Dealbot performs direct `/piece/{pieceCid}` retrieval from the SP and confirms the bytes match the pieceCid. | Yes | [`pull-check.service.ts`](../../apps/backend/src/pull-check/pull-check.service.ts) |
```
BigLep:
```suggestion
| <a id="pullRequestIntegrityChecked"></a>`pullRequestIntegrityChecked` | Dealbot completed direct `/piece/{pieceCid}` retrieval from the SP and confirmed the bytes match the pieceCid. | Yes | [`pull-check.service.ts`](../../apps/backend/src/pull-check/pull-check.service.ts) |
```

docs/checks/events-and-metrics.md
```diff
@@ -87,6 +121,10 @@ sequenceDiagram
...
+| <a id="pullRequestStartedMs"></a>`pullRequestStartedMs` | Pull | [`pullRequestSubmittedToSp`](#pullRequestSubmittedToSp) | [`pullRequestStartedBySp`](#pullRequestStartedBySp) | Time from `pullPieces` submission to the SP reading the first byte of `/api/piece/{pieceCid}`. Skipped (no observation) when the SP never fetches from dealbot. | [`pull-check.service.ts`](../../apps/backend/src/pull-check/pull-check.service.ts), [`pull-piece.controller.ts`](../../apps/backend/src/pull-check/pull-piece.controller.ts) |
+| <a id="pullRequestCompletionLatencyMs"></a>`pullRequestCompletionLatencyMs` | Pull | [`pullRequestSubmittedToSp`](#pullRequestSubmittedToSp) | [`pullRequestIsTerminal`](#pullRequestIsTerminal) | Time from `pullPieces` submission to terminal SP pull status. Observed once on success and once on failure. | [`pull-check.service.ts`](../../apps/backend/src/pull-check/pull-check.service.ts) |
```
BigLep:
```suggestion
| <a id="pullRequestCompletionLatencyMs"></a>`pullRequestCompletionLatencyMs` | Pull | [`pullRequestSubmittedToSp`](#pullRequestSubmittedToSp) | [`pullRequestIsTerminal`](#pullRequestIsTerminal) | Time from `pullPieces` submission to terminal SP pull status. Emitted once for the check, either on success or failure. | [`pull-check.service.ts`](../../apps/backend/src/pull-check/pull-check.service.ts) |
```

Tying to make clearer that there is only one data point emitted here.

docs/checks/pull-check.md
```diff
@@ -0,0 +1,156 @@
...
+
+This document is the **source of truth** for how dealbot's Pull check works.
+
+Source code links throughout this document point to the current implementation.
+
+For event and metric definitions used by the dashboard, see [Dealbot Events & Metrics](./events-and-metrics.md).
+
+## Overview
+
+A "pull check" exercises the **storage provider pull-to-park pathway**: dealbot publishes a temporary piece at `/api/piece/{pieceCid}`, asks the SP to fetch (pull) and park it via the Synapse `pullPieces` API, waits for a terminal SP pull status, and finally re-fetches the piece from the SP to verify byte-for-byte integrity.
```
BigLep:
```suggestion
A "pull check" exercises the **[storage provider pull-to-park pathway](https://docs.filecoin.cloud/developer-guides/storage/upload-pipeline/#pull-phase-sp-to-sp-transfer)**: dealbot publishes a temporary piece at `/api/piece/{pieceCid}`, asks the SP to fetch (pull) and park it via the Synapse `pullPieces` API, waits for a terminal SP pull status, and finally re-fetches the piece from the SP to verify byte-for-byte integrity.
```

docs/checks/pull-check.md
BigLep: Ok, that's a good detail to maybe add to this section? Basically when we make the `pullPieces` call, we pass Dealbot as the `payee` to pass Curio's validation... ?

BigLep (approved):
Thanks for making the updates. I had a couple of small comments. Feel free to merge after reviewing and incorporating where appropraite.

BigLep (approved):
Let's merge!

Issue #379, opened 2026-03-19
StorageManager context cache not invalidated after dataset termination


beck-8:
https://telemetry.betterstack.com/dashboards/1eH3gR?rf=now-72h&rt=now&vs%5BproviderId%5D=6

I'm not sure this should be made compatible by dealbot. Or synapse-sdk does cache invalidation?
```
{"type":"error","name":"Error","message":"StorageContext addPieces failed: Failed to add piece to data set - Failed to add pieces.\n\nDetails: Service Provider PDP\nData set has been terminated due to unrecoverable proving failure\n","stack":"Error: StorageContext addPieces failed: Failed to add piece to data set - Failed to add pieces.\n\nDetails: Service Provider PDP\nData set has been terminated due to unrecoverable proving failure\n\n at createError (file:///app/node_modules/.pnpm/@filoz+synapse-sdk@0.36.1_iso-filecoin@7.4.7_typescript@5.9.3__typescript@5.9.3_zod@4.3.6/node_modules/@filoz/synapse-sdk/dist/src/utils/errors.js:9:22)\n at StorageContext._processPendingPieces (file:///app/node_modules/.pnpm/@filoz+synapse-sdk@0.36.1_iso-filecoin@7.4.7_typescript@5.9.3__typescript@5.9.3_zod@4.3.6/node_modules/@filoz/synapse-sdk/dist/src/storage/context.js:515:32)\n at process.processTicksAndRejections (node:internal/process/task_queues:104:5
```

BigLep:
2026-04-22 discussion: yes, we should make sure data storage and retrieval checks are more resilient to state mismatch between onchain and dealbots db.

If a dataset or piece is deleted onchain, we shouldn't try to do a data storage or retrieval check. (We need to see if this is something going on in synapse or within dealbot itself.)

In terms of keeping database and onchain state syncrhonized, that is a separate item. We'll track that work separately in #465.

PR #505, opened 2026-05-04
fix(deal): cancel onStored addons when upload fails

SgtPooki:
## What changed

`StorageContext.upload()` in `@filoz/synapse-sdk` fires the `onStored` callback before `commit()`/`addPieces`, so dealbot's IPNI monitoring (started inside `onStored`) runs detached from the main upload flow. When `executeUpload` throws afterward (e.g. a 409 on `POST /pdp/data-sets/{id}/pieces` against a Curio-terminated dataset), the deal already failed but the IPNI poll keeps running for its 120s timeout and emits a misleading `ipni_tracking_failed` event.

This wires an `AbortController` for the detached addons, composed with the parent signal via `AbortSignal.any`. On upload failure the `finally` block aborts and drains the pending addon promise. IPNI's existing `signal?.throwIfAborted()` guards (`ipni.strategy.ts:255`, `deal-addons.service.ts:193`) suppress the spurious failure log/metric.

Closes #503.

## How to verify

```
cd apps/backend && pnpm test -- --run src/deal/deal.service.spec.ts
```

New regression test: `aborts in-flight onStored addons when executeUpload fails after onStored fires`.

## Notes

Underlying terminated-dataset reuse (#379) is unaffected. The race window where the SDK reuses a Curio-terminated dataset still happens; this fix bounds blast radius (no leaked 120s poll, no spurious metric) but does not prevent the first failure.

### DIFF
--- a/apps/backend/src/deal/deal.service.ts
+++ b/apps/backend/src/deal/deal.service.ts
@@ -226,6 +226,7 @@ export class DealService implements OnModuleInit, OnModuleDestroy {
     const addonSignal: AbortSignal = signal ? AbortSignal.any([signal, addonAbortCtrl.signal]) : addonAbortCtrl.signal;
     /** Wrapper object so TS preserves the union type across closure mutation. */
     const onStoredAddons: { promise: Promise<boolean> | null } = { promise: null };
+    let addonsAwaited = false;
     let storedError: Error | undefined;
 
     try {
@@ -436,6 +437,7 @@ export class DealService implements OnModuleInit, OnModuleDestroy {
       // wait for onStored handlers to complete
       if (onStoredAddons.promise != null) {
         const storedOk = await onStoredAddons.promise;
+        addonsAwaited = true;
         if (!storedOk) {
           throw storedError ?? new Error("Upload completion handlers failed");
         }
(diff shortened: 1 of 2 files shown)


silent-cipher (approved):
looks good to me

apps/backend/src/deal/deal.service.ts
```diff
@@ -529,6 +533,15 @@ export class DealService implements OnModuleInit, OnModuleDestroy {
 
       throw error;
     } finally {
+      if (onStoredAddons.promise != null) {
+        const pending = onStoredAddons.promise;
+        addonAbortCtrl.abort();
```
silent-cipher, line 540: nit: `addonAbortCtrl.abort()` always fires, even on the happy path

silent-cipher (approved):

apps/backend/src/deal/deal.service.ts
SgtPooki: fixed with [`f2820ec` (this PR)](https://github.com/FilOzone/dealbot/pull/505/commits/f2820ec8302fc09d8028238673e53e0a1926a529)

PR #518, opened 2026-05-08
fix: handle PDP-terminated datasets via data_set_creation repair

### LINKED ISSUE #379: StorageManager context cache not invalidated after dataset termination

SgtPooki:
## What changed

`data_set_creation` classifies each provider's data set slot as `missing | live | terminated` and runs a bounded repair on `terminated` instead of failing later from the deal path.

### Repair sequence (`data_set_creation`, terminated path)

1. `synapse.storage.terminateDataSet({ dataSetId })`. Already-terminated reverts treated as no-ops (idempotent).
2. Poll `warmStorageService.getDataSet({ dataSetId }).pdpEndEpoch != 0n` with exponential backoff capped at 8s, 60s ceiling.
3. Mark every `Deal` row with that `dataSetId` as `cleaned_up = true, cleaned_up_at = now()` in one transaction.
4. Return without provisioning a replacement this tick. The next tick sees `missing` and provisions.

### Terminated handling on the deal path

`handleDealJob` probes the baseline metadata up front, plus the selected `dealbotDS` index when `minNumDataSetsForChecks > 1`. `createDeal` repeats the probe after `createContext`, covering the `triggerDeal` (dev-tools) path which does not run the upfront check. All three probe sites throw `DealJobTerminatedDataSetError` (in `apps/backend/src/common/errors.ts`) on terminated. `handleDealJob`'s outer catch maps the typed error to `handler_result: "error"` and the log event `deal_job_failed_terminated_dataset`. `triggerDeal` background catches the same error and marks the placeholder `Deal` row `FAILED` with an `errorMessage` describing the deferral.

## How to verify

`pnpm test` from `apps/backend` (342 tests pass). Coverage: repair handler, both upfront terminated probes asserting `handler_result: "error"`, the in-`createDeal` guard, idempotent repair, `getDataSetProvisioningStatus` for all three states, and `isDataSetLive` error classification.

## Notes / risks

A terminated data set records `handler_result: "error"` in Prometheus but resolves the pg-boss worker callback, so pg-boss sees the job as completed (not failed). With `retryLimit: 0` this has no behavioral impact; the next deal job arrives on the normal schedule tick.

Upstream trackers do not gate this PR: FilOzone/synapse-sdk#780 (SDK liveness check + typed error on write paths), FilOzone/filecoin-services#473 (PDP → FWSS termination sync).

Closes #379.


apps/backend/src/dev-tools/dev-tools.service.ts
```diff
@@ -160,12 +160,20 @@ export class DevToolsService {
         },
       });
 
-      this.logger.log({
-        ...dealLogContext,
-        event: "background_deal_completed",
-        message: "Background deal completed successfully",
-        pieceCid: deal.pieceCid,
-      });
+      if (deal === null) {
```
silent-cipher: Wouldn't this leave an orphaned `PENDING` row on skip?

silent-cipher (changes requested):
This looks good to me. I just have one comment and one suggestion that might be worth considering.

Suggestion:

We already have `getDataSetProvisioningStatus`, which combines `checkDataSetExists` and `isDataSetLive`. Right now, we run `checkDataSetExists` at the start of the deal job and later run `isDataSetLive` inside `createDeal`. It seems we could instead call `getDataSetProvisioningStatus` upfront at the start of the deal job and avoid splitting the validation across multiple places.

This would also simplify triggerDeal, since we could validate everything before persisting the `PENDING` row.

apps/backend/src/dev-tools/dev-tools.service.ts
SgtPooki: fixed in [`98dc554` (this PR)](https://github.com/FilOzone/dealbot/pull/518/commits/98dc554f299e469872e915e1e6e80432a647a311)

apps/backend/src/deal/deal.service.spec.ts
```diff
@@ -1167,37 +1207,199 @@ describe("DealService", () => {
...
+      const result = await service.repairTerminatedDataSet("0xaaa", 9n, undefined, 5_000);
+
+      expect(terminateMock).toHaveBeenCalledWith({ dataSetId: 9n });
+      expect(waitForReceiptMock).toHaveBeenCalledWith({ hash: "0xhash" });
+      expect(updateFn).toHaveBeenCalledWith(
+        { dataSetId: 9n, cleanedUp: false },
+        expect.objectContaining({ cleanedUp: true, cleanedUpAt: expect.any(Date) }),
+      );
+      expect(result.dealsAffected).toBe(2);
+      expect(result.pdpEndEpoch).toBe(12345n);
```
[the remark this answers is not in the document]
SgtPooki: test runs for 8ms.. ignoring.

silent-cipher (approved):
This looks good to me.

What I meant by moving the check upfront was something closer to - 

```
if minDataSets > 1 && dsIndex > 0:
    status = getDataSetProvisioningStatus(dsIndex slot)
    if live → use dsIndex slot (no baseline check needed — we're not using it)
    if missing or terminated → fall back: check baseline
else:
    check baseline

baseline check:
    if terminated → throw DealJobTerminatedDataSetError
    if missing or live → proceed (data_set_creation owns repair/provisioning)
```

With this approach, we could likely remove the `isDataSetLive` guard entirely by moving `getDataSetProvisioningStatus` into `triggerDeal`. I assume the current `isDataSetLive` guard inside `createDeal` exists because the provisioning validation is currently split across multiple stages.

This would simplify the overall flow and also avoid one extra `createContext` + `validateDataSet` call per job.

PR #519, opened 2026-05-08
fix: stop retention counter double-counts

SgtPooki:
## What changed

Data retention polls now reload persisted baselines at the start of every run and treat Postgres as the shared source of truth across worker pods. Provider processing computes deltas without mutating local counter state; the service persists each new baseline first, then updates memory and increments `dataSetChallengeStatus` only after that provider write succeeds.

This fixes #517, where sequential `data_retention_poll` jobs could run on different worker pods and reuse stale process-local baselines, inflating summed Prometheus counters.

The data-retention docs now describe the cross-pod baseline lifecycle and persist-before-counter order.

## How to verify

- `pnpm -C apps/backend test src/data-retention/data-retention.service.spec.ts`
- `pnpm -C apps/backend typecheck`
- `pnpm -C apps/backend check:ci`

## Notes

Provider baseline writes still run in parallel within each batch. `data_retention_poll` is enqueued as a pg-boss global singleton by job type; this PR also skips overlapping `pollDataRetention()` calls inside one worker process.

I confirmed the new pod-alternation regression test fails against the old service with `[10, 10]` increments and passes on this branch with `[5, 5]`.

### DIFF
--- a/apps/backend/src/data-retention/data-retention.service.spec.ts
+++ b/apps/backend/src/data-retention/data-retention.service.spec.ts
@@ -32,6 +32,12 @@ const makeProvider = (overrides: Partial<ProviderEntry> = {}): ProviderEntry => 
 
 describe("DataRetentionService", () => {
   let service: DataRetentionService;
+  let baselineRows: Array<{
+    providerAddress: string;
+    faultedPeriods: string;
+    successPeriods: string;
+    lastBlockNumber: string;
+  }>;
   let configServiceMock: ConfigService<IConfig, true>;
   let walletSdkServiceMock: {
     getTestingProviders: ReturnType<typeof vi.fn>;
@@ -116,12 +122,26 @@ describe("DataRetentionService", () => {
       remove: vi.fn(),
     };
 
+    baselineRows = [];
     mockBaselineRepository = {
-      find: vi.fn().mockResolvedValue([]),
-      upsert: vi.fn().mockResolvedValue(undefined),
-      delete: vi.fn().mockResolvedValue(undefined),
+      find: vi.fn().mockImplementation(async () => baselineRows.map((row) => ({ ...row }))),
+      upsert: vi.fn().mockImplementation(async (row: (typeof baselineRows)[number]) => {
+        const index = baselineRows.findIndex(
+          (existing) => existing.providerAddress.toLowerCase() === row.providerAddress.toLowerCase(),
+        );
+        if (index >= 0) {
+          baselineRows[index] = { ...baselineRows[index], ...row };
+        } else {
+          baselineRows.push({ ...row });
+        }
+      }),
+      delete: vi.fn().mockImplementation(async ({ providerAddress }: { providerAddress: string }) => {
+        baselineRows = baselineRows.filter(
+          (row) => row.providerAddress.toLowerCase() !== providerAddress.toLowerCase(),
+        );
+      }),
     };
-    mockSPRepository = { find: vi.fn() };
+    mockSPRepository = { find: vi.fn().mockResolvedValue([]) };
     const clickhouseServiceMock = { insert: vi.fn(), probeLocation: "test" } as unknown as ClickhouseService;
     service = new DataRetentionService(
       configServiceMock,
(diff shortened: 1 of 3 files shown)


apps/backend/src/data-retention/data-retention.service.ts
```diff
@@ -131,38 +146,38 @@ export class DataRetentionService {
...
-                message: "Failed to persist baseline to database",
-                error: toStructuredError(result.reason),
-              });
-            }
-          }
+              try {
+                await this.persistBaseline(result.value.providerAddress, result.value.baseline, blockNumberBigInt);
+                this.applyPersistedProviderResult(result.value);
+              } catch (error) {
+                hasProcessingErrors = true;
```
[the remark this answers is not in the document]
SgtPooki: Fixed: split baseline persistence and result application into separate error paths. Persist failures still skip apply; apply failures now log `provider_result_apply_failed`.

apps/backend/src/data-retention/data-retention.service.spec.ts
```diff
@@ -116,12 +122,24 @@ describe("DataRetentionService", () => {
...
+          (existing) => existing.providerAddress.toLowerCase() === row.providerAddress.toLowerCase(),
+        );
+        if (index >= 0) {
+          baselineRows[index] = { ...baselineRows[index], ...row };
+        } else {
+          baselineRows.push({ ...row });
+        }
+      }),
+      delete: vi.fn().mockImplementation(async ({ providerAddress }: { providerAddress: string }) => {
+        baselineRows = baselineRows.filter((row) => row.providerAddress !== providerAddress);
```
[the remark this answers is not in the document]
SgtPooki: Fixed: the in-memory baseline stub now deletes case-insensitively to match upsert behavior.

apps/backend/src/data-retention/data-retention.service.ts
```diff
@@ -16,31 +16,38 @@ import { type ProviderDataSetResponse } from "../pdp-subgraph/types.js";
...
+};
+
 @Injectable()
 export class DataRetentionService {
   private readonly logger = new Logger(DataRetentionService.name);
 
   private static readonly MAX_PROVIDER_BATCH_LENGTH = 50;
   // NOTE: taken from https://github.com/FilOzone/filecoin-services/blob/c04be93aa680082e359481f0776e41ed157a2ac2/service_contracts/src/FilecoinWarmStorageService.sol#L26
   private static readonly CHALLENGES_PER_PROVING_PERIOD = 5n;
+  private pollInProgress = false;
```
silent-cipher: I'm not fully following the need for `pollInProgress`. What are the chances of the data-retention job being picked up concurrently by the same worker?

apps/backend/src/data-retention/data-retention.service.ts
```diff
@@ -16,31 +16,38 @@ import { type ProviderDataSetResponse } from "../pdp-subgraph/types.js";
...
-    string,
-    {
-      faultedPeriods: bigint;
-      successPeriods: bigint;
-    }
-  >;
-
-  /** Whether baselines have been loaded from the database */
-  private baselinesLoaded = false;
+  private readonly providerCumulativeTotals: Map<string, ProviderBaseline>;
```
silent-cipher: Is there still a use case for this after the patch? It seems like it may now only add additional maintenance burden.

silent-cipher (changes requested):
Overall looks good. Just two comments.

apps/backend/src/data-retention/data-retention.service.ts
SgtPooki: yea, you're totally right. will never happen, good catch.

apps/backend/src/data-retention/data-retention.service.ts
SgtPooki: eh, there was a use for it, but it can be removed if we pass it around.

SgtPooki: fixed in [`8b813bc` (this PR)](https://github.com/FilOzone/dealbot/pull/519/commits/8b813bc904de2fed40b560c64176a043a24fbab9)

apps/backend/src/data-retention/data-retention.service.ts
SgtPooki: fixed in [`8b813bc` (this PR)](https://github.com/FilOzone/dealbot/pull/519/commits/8b813bc904de2fed40b560c64176a043a24fbab9)

silent-cipher (approved):
LGTM!

PR #521, opened 2026-05-11
chore: update Synapse stack for filecoin-pin 0.21

rjan90:
- Update Synapse dependencies to:
  - `@filoz/synapse-core` 0.5.2
  - `@filoz/synapse-sdk` 0.41.0
  - `filecoin-pin` ^0.21.0
- Align `viem` to ^2.48.11 so Synapse peer types resolve consistently.
- Update Dealbot upload progress handling for the filecoin-pin 0.21 event rename:
  - `onStored` -> `stored`
  - `onPiecesAdded` -> `piecesAdded`
  - `onPiecesConfirmed` -> `piecesConfirmed`
- Refresh related tests and ClickHouse schema comments.

Closes #516

PR #526, opened 2026-05-12
feat: filter out sps with dev tags

### LINKED ISSUE #454: filter out SPs with dev or other known "not useful to test" tags
> I was going to say that we should just pull `serviceStatus` out of the provider info and do the filtering here because it's still there, Synapse just stopped caring about it. But it turns out we're actively stripping non-official fields from the capabilities list in Synapse now so you can't. Here's a fix that brings it back:  https://github.com/FilOzone/synapse-sdk/pull/687
> 
> Then here, in apps/backend/src/wallet-sdk/wallet-sdk.service.ts `loadProviders` you do the filtering, comparing `info.pdp.extraCapabilities?.serviceStatus` to `'0x646576'` (`"dev"`), something like this with bonus logging:
> 
> ```diff
> -      const validProviders = providerInfos.filter((info) => !!info);
> +      const validProviders = providerInfos.filter((info) => {
> +        if (!info) return false;
> +        // Skip providers that have declared serviceStatus=dev in their capabilities
> +        const serviceStatus = info.pdp.extraCapabilities?.serviceStatus;
> +        if (serviceStatus === "0x646576") { // "dev"
> +          this.logger.log({
> +            event: "provider_skip_dev",
> +            message: `Skipping dev provider "${info.name}" (ID ${info.id})`,
> +            providerId: info.id,
> +            providerName: info.name,
> +          });
> +          return false;
> +        }
> +        return true;
> +      });
> ```
> 
> This is not to argue that we shouldn't also have a way to opt-out on our end, but it would be nice for an SP to signal without having to tell us that they don't want to be tested. I know Zen's going to be setting up his test SP again so will be in the same boat probably. 

 _Originally posted by @rvagg in [#364](https://github.com/FilOzone/dealbot/issues/364#issuecomment-4107733425)_

silent-cipher:
closes #454 

**Note** - This PR is built upon #521 .


apps/backend/src/wallet-sdk/wallet-sdk.service.ts
```diff
@@ -395,6 +407,21 @@ export class WalletSdkService implements OnModuleInit {
...
+      const lower = k.toLowerCase();
+      return lower === "servicestatus" || lower === "service_status";
+    });
+    if (key === undefined) return false;
+    try {
+      return hexToString(extraCaps[key]).toLowerCase() === "dev";
+    } catch {
+      return false;
+    }
+  }
```
SgtPooki, line 414: we can probably simplify this greatly by just checking that `serviceStatus === '0x646576'`.

I don't think we should get to a place where we're allowing different variations of "serviceStatus" keys.

apps/backend/src/wallet-sdk/wallet-sdk.service.spec.ts
```diff
@@ -265,4 +266,98 @@ describe("WalletSdkService", () => {
       await expect(service.ensureWalletAllowances()).rejects.toThrow();
     });
   });
+
+  describe("isDevProvider", () => {
```
SgtPooki, line 270: we should reduce the number of tests here, we don't need all these.

SgtPooki (changes requested):
two comments, and we should probably also add a short doc for SPs who may not want to be tested by dealbot since there is no other documented convention for this. Since there is no convention, it might make sense to try to check all the things, but i think we can define the contract here in dealbot and just say "set this to that if you don't want to be considered for dealbot testing or Approval status eligibility" or similar

@BigLep what do you think?

BigLep:
> i think we can define the contract here in dealbot and just say "set this to that if you don't want to be considered for dealbot testing or Approval status eligibility"

That sounds good to me. Basically I think we should add an FAQ entry to https://github.com/FilOzone/dealbot/blob/main/docs/checks/production-configuration-and-approval-methodology.md#appendix

apps/backend/src/wallet-sdk/wallet-sdk.service.ts
silent-cipher: I found out that kuba is using `service_status === '0x646576'`

silent-cipher: And another reason for checking all the cases was there is no official contract defined for `serviceStatus`.

apps/backend/src/wallet-sdk/wallet-sdk.service.spec.ts
silent-cipher: [22b6808](https://github.com/FilOzone/dealbot/pull/526/commits/22b6808918dad87171e9b7fba8d84d2da2ab731e)

apps/backend/src/wallet-sdk/wallet-sdk.service.ts
silent-cipher: I didn’t simplify this because there’s an existing SP using `service_status`. Let me know if you’d still prefer that I simplify it.

silent-cipher: I've simplified to dealbot documented conventions in [46d3610](https://github.com/FilOzone/dealbot/pull/526/commits/46d3610917352d4e207fb62811c2fa65ea0366b4)

BigLep (commented):
Doc updates look good to me. Will leave to @SgtPooki to give the approval.

SgtPooki (approved):

PR #535, opened 2026-05-13
refactor: centralize data-set probe in DealService

SgtPooki:
## What changed

Lift the dsIndex selection + `getDataSetProvisioningStatus` probe out of `handleDealJob` into a new `DealService.resolveDataSetMetadataForDeal`, invoked from `createDealForProvider`. `handleDealJob` delegates to `createDealForProvider` and maps `DealJobTerminatedDataSetError` to `handler_result: "error"`. `triggerDeal` (dev-tools) inherits the same policy automatically.

Follow-up to #518 review feedback at https://github.com/FilOzone/dealbot/pull/518#pullrequestreview-4282036395.

### Behavior changes

- When `minNumDataSetsForChecks > 1` and the randomly selected indexed slot is PDP-terminated, the deal job falls back to the baseline slot instead of failing the whole job. The terminated index is logged (`deal_job_dataset_index_terminated`); `data_set_creation` still owns repair.
- Common case (`minDataSets > 1`, indexed slot live) drops from 2 `getDataSetProvisioningStatus` calls to 1.

### Why this differs from silent-cipher's exact suggestion

Silent-cipher also suggested removing the inner `isDataSetLive` guard inside `createDeal` (deal.service.ts:266-272). This PR keeps it. Reasoning:

- The upfront probe and `createDeal`'s own `createContext` are separated by `prepareDealInput` + `prepareUploadPayload` (CAR build, random data fetch, which takes seconds). PDP termination is async on-chain state; the window is wide enough to reproduce #379 without the guard.
- `triggerDeal` runs deal creation in a detached background promise, which widens the race further on the dev-tools path.
- The inner guard probes the exact `dataSetId` returned by the upload's own `createContext`, the only place that `dataSetId` is known. Upfront probes are advisory; the inner guard is the commit-time check.
- Cost is one `validateDataSet` RPC. Cost of failure is a wasted upload plus #379 recurrence. Asymmetric, favor the guard.

Four peer reviewers (Cursor, Claude, Codex, Gemini) ran on the proposed change. Three of four (Cursor, Claude, Codex) recommended keeping the inner guard and centralizing the upfront policy in `createDealForProvider`. This PR implements that consensus.

## How to verify

`pnpm test` from repo root. 345 tests pass (was 342 on #518; net +3 covering all `resolveDataSetMetadataForDeal` branches: minDataSets=1 live/terminated, indexed live, indexed missing fallback, indexed terminated fallback, indexed throw fallback, index 0, signal abort).

## Notes / risks

- `createDealForProvider`'s `extraDataSetMetadata` option is removed (caller no longer computes metadata). No external consumers exist outside `apps/backend`.
- Event `deal_job_dataset_index_terminated` is now emitted by `DealService` (was `JobsService`). Same event name and shape; dashboards keyed on the event name continue to work.

### DIFF
--- a/apps/backend/src/deal/deal.service.spec.ts
+++ b/apps/backend/src/deal/deal.service.spec.ts
@@ -1252,6 +1252,132 @@ describe("DealService", () => {
     });
   });
 
+  describe("resolveDataSetMetadataForDeal", () => {
+    const providerInfo: PDPProviderEx = {
+      id: 101n,
+      serviceProvider: "0xprovider",
+      payee: "0x100",
+      name: "Test Provider",
+      description: "Test Provider",
+      isActive: true,
+      isApproved: true,
+      pdp: {
+        serviceURL: "service url",
+        minPieceSizeInBytes: 0n,
+        maxPieceSizeInBytes: 100n,
+        storagePricePerTibPerDay: 1n,
+        minProvingPeriodInEpochs: 1n,
+        location: "location",
+        paymentTokenAddress: "0x100",
+        ipniPiece: true,
+        ipniIpfs: true,
+      },
+    };
+
+    let probeSpy: ReturnType<typeof vi.spyOn>;
+
+    beforeEach(() => {
+      vi.spyOn(mockWalletSdkService, "getProviderInfo").mockReturnValue(providerInfo);
+      probeSpy = vi.spyOn(service, "getDataSetProvisioningStatus");
+    });
+
+    afterEach(() => {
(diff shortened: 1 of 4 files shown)


silent-cipher (approved):
I still don’t fully buy the TOCTOU distinction here 😁 . But,  “Cost is one validateDataSet RPC. Cost of failure is a wasted upload plus...” was persuasive enough for me to accept that `isDataSetLive` probably exists for a good reason.

PR #538, opened 2026-05-14
feat(retrieval): parallel IPNI + transport with retrievalTransportStatus counter

SgtPooki:
## What changed

In pgBoss mode, \`performAllRetrievals\` gated the \`/ipfs\` transport on IPNI verification: any IPNI lag bypassed the transport path entirely and pinned \`retrievalStatus=failure.*\`. Combined with the silent-timeout bug (#524), the failure mode was invisible in logs.

This PR runs IPNI and transport concurrently and composes the terminal status as \`AND(discoverabilityStatus, retrievalTransportStatus)\`. A new \`retrievalTransportStatus\` counter records the \`/ipfs\`-only outcome so a single dashboard panel can attribute a \`retrievalStatus\` failure to IPNI vs transport.

Implements option A from #524. Matches the parallel flow already documented in \`docs/checks/retrievals.md\`.

## Files

- \`apps/backend/src/retrieval/retrieval.service.ts\` — parallel orchestration via \`Promise.allSettled\`, classify + compose helpers.
- \`apps/backend/src/metrics-prometheus/{metrics-prometheus.module.ts,check-metrics.service.ts}\` — register \`retrievalTransportStatus\` counter and \`recordTransportStatus\` method.
- \`docs/checks/{events-and-metrics.md,retrievals.md}\` — document the new counter and the composite-status rule.
- \`apps/backend/src/retrieval/retrieval.service.spec.ts\` — 4 new tests covering IPNI+transport success, IPNI fail / transport success, IPNI timeout / transport success, and concurrent execution ordering.

## How to verify

\`\`\`
pnpm --filter dealbot-backend typecheck
pnpm --filter dealbot-backend lint
pnpm --filter dealbot-backend test
\`\`\`

After deploy, watch:
- \`retrievalTransportStatus{value=success}\` should equal \`/ipfs\` success volume per SP.
- For approved SPs with high \`retrievalStatus=failure.*\` but high \`retrievalTransportStatus=success\`, the failure is in IPNI discoverability, not transport.

## Notes

\`retrievalStatus\` semantics unchanged (AND of discoverability and transport). No dashboard migration needed; the new \`retrievalTransportStatus\` chart is opt-in.

Observability follow-ups for the silent-timeout cases (the catch-order bug in \`ipni-verification.service.ts\` and the missing \`retrieval_skipped_ipni_failure\` log) ship as a separate PR (option C from #524) on top of this one.


docs/checks/events-and-metrics.md
BigLep, line 1: Per https://github.com/FilOzone/dealbot/blob/main/docs/checks/data-storage.md#deal-status-progression and https://github.com/FilOzone/dealbot/blob/main/docs/checks/events-and-metrics.md#retrievalStatus, this `retrievalStatus` counter is supposed to increase when the `ipfsRetrievalIntegrityChecked` event fires. It wasn't intended to be when `ipfsRetrievalIntegrityChecked` and `ipniVerificationComplete`.

Assuming we implement it that way, do we even need a counter for when `ipfsRetrievalIntegrityChecked` and `ipniVerificationComplete` occur? Back when specing this out, I wasn't seeing what that got us.  

My understanding of the situation:
* Today: `retrievalStatus` is successful when `ipfsRetrievalIntegrityChecked` and `ipniVerificationComplete`occur with success, and those happen sequentially.
* Your proposal: 
  - `retrievalStatus` is successful when `ipfsRetrievalIntegrityChecked` and `ipniVerificationComplete` occur with success, and those happen in parallel.
  - Add `retrievalTransportStatus` for when `ipfsRetrievalIntegrityChecked` fires.
* My proposal (as originally speced, or at least what was meant but I can accept it wasn't clear)
  -   `retrievalStatus` when `ipfsRetrievalIntegrityChecked` fires.
  - Discoverability and Retrieval portions happen in parallel.
  - No additional counters.

docs/checks/events-and-metrics.md
```diff
@@ -102,7 +102,8 @@ sequenceDiagram
...
+| <a id="retrievalStatus"></a>`retrievalStatus` | Data Storage, Retrieval | [`ipfsRetrievalIntegrityChecked`](#ipfsRetrievalIntegrityChecked) | `success`, `failure.timedout`, `failure.other` from [Data Storage Sub-status meanings](./data-storage.md#sub-status-meanings). Composite: `success` requires both `discoverabilityStatus=success` and `retrievalTransportStatus=success`. | [`retrieval.service.ts`](../../apps/backend/src/retrieval/retrieval.service.ts) |
+| <a id="retrievalTransportStatus"></a>`retrievalTransportStatus` | Retrieval | When the `/ipfs` transport stage completes (success, all-blocks fetched and validated; or failure on HTTP, content, or abort). Scoped to the transport stage; does **not** include IPNI discoverability. Pair with [`discoverabilityStatus`](#discoverabilityStatus) to attribute a `retrievalStatus` failure to IPNI vs transport. | `success`, `failure.timedout`, `failure.other` from [Data Storage Sub-status meanings](./data-storage.md#sub-status-meanings). | [`retrieval.service.ts`](../../apps/backend/src/retrieval/retrieval.service.ts) |
```
BigLep: Per file comment, I'm questioning whether we need to change anything here. I think we can just implement `retrievalStatus` as originally specced aand just have one counter here.

docs/checks/retrievals.md
```diff
@@ -54,7 +54,7 @@ Source: [`retrieval.service.ts` (`selectRandomDealsForRetrieval`)](../../apps/ba
 
 ### Retrieval Checks
 
-For each selected piece, dealbot performs the following in parallel:
+For each selected piece, dealbot performs the following in parallel. Each stage emits its own sub-status counter ([`discoverabilityStatus`](./events-and-metrics.md#discoverabilityStatus) for IPNI, [`retrievalTransportStatus`](./events-and-metrics.md#retrievalTransportStatus) for `/ipfs`). The composite [`retrievalStatus`](./events-and-metrics.md#retrievalStatus) records `success` only when both sub-statuses succeed.
```
BigLep: I'm not sure more documentation is needed here, but if we are wanting to add extra documentation around metrics, lets put it in https://github.com/FilOzone/dealbot/blob/main/docs/checks/retrievals.md#metrics-recorded ?

BigLep (dismissed):
Requesting changes to get aligned on what counters are emitted and when.

apps/backend/src/retrieval/retrieval.service.ts
```diff
@@ -130,61 +131,108 @@ export class RetrievalService {
...
+        ? this.verifyIpniForRetrieval(
+            ipniContext as { rootCid: CID; blockCids: CID[] },
+            deal.id,
+            provider,
+            providerLabels,
+            signal,
+          )
+        : Promise.resolve({ ok: true as const });
+
+      const [transportSettled, ipniSettled] = await Promise.allSettled([transportPromise, ipniPromise]);
```
silent-cipher, line 145: This may extend runtime to the slower of the two timeouts. For example, we could end up waiting for the IPNI timeout even though the transport has already implied `failure.timedout` since there’s no early exit.

silent-cipher (commented):
code portion looks good overall. Just one comment.

apps/backend/src/retrieval/retrieval.service.ts
SgtPooki: good catch

docs/checks/events-and-metrics.md
SgtPooki: yea makes sense.. I was overthinking this one

apps/backend/src/retrieval/retrieval.service.ts
SgtPooki: yea, i think this is actually what we want here tbh. if we abort ipni when transport fails it gets recorded as `failure.timedout` (because that's what a signal abort looks like) instead of whatever ipni was actually going to return. that corrupts discoverabilityStatus.

also we have an invariant we care about: transport fail should imply ipni fail. running them independently lets us alert on the violation case (ipni=success while transport=fail), which under the invariant should basically never happen. if we early-exit ipni we can't see that anymore.

the job timeout is the ceiling and will mark anything still in-flight as failure.timedout when it fires. inner stage timeouts (ipni verify, transport per-request) are meant to be independent budgets. so worst case runtime when one side fails is `max(ipniTimeout, transportTimeout)`, bounded by the job ceiling. seems like the right tradeoff for the signal we get.

apps/backend/src/retrieval/retrieval.service.ts
```diff
@@ -130,86 +131,39 @@ export class RetrievalService {
...
 
-      // If this catch block fires, it's either:
-      // 1. The job timeout fired (signal aborted) - record failure.timedout once
-      // 2. A catastrophic error occurred (provider not found, etc.) - re-throw for logging
-      // 3. Individual HTTP timeouts are captured by Promise.allSettled in testAllRetrievalMethods
-      //    and converted to FAILED records through the normal flow
-      caughtError = error;
+      // retrievalStatus is scoped to the transport stage (ipfsRetrievalIntegrityChecked).
+      // IPNI outcomes are recorded independently via discoverabilityStatus.
+      terminalStatus = this.classifyTransportOutcome(transportSettled, signal);
```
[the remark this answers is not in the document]
SgtPooki: not a contradiction. per data-storage.md deal status progression (https://github.com/FilOzone/dealbot/blob/main/docs/checks/data-storage.md#deal-status-progression), retrievalStatus is one of four sub-statuses (Upload, Onchain, Discoverability, Retrieval). IPNI feeds discoverabilityStatus, transport feeds retrievalStatus. composite is dataStorageStatus, not retrievalStatus.

retrievals.md was ambiguous because it listed all assertions in one table without attribution. just clarified in the table to show which sub-status each assertion feeds.

apps/backend/src/retrieval/retrieval.service.spec.ts
```diff
@@ -323,6 +323,234 @@ describe("RetrievalService timeouts", () => {
...
+      failedCIDs: [{ cid: "x", reason: "missing" }],
+      verifiedAt: new Date().toISOString(),
+    });
+    mockRetrievalAddonsService.testAllRetrievalMethods.mockResolvedValue(successfulTransport);
+
+    await service.performAllRetrievals(buildDealWithIpni());
+
+    expect(mockRetrievalMetrics.recordStatus).toHaveBeenCalledWith(labels, "success");
+    expect(mockDiscoverabilityMetrics.recordStatus).toHaveBeenCalledWith(labels, "failure.other");
+  });
```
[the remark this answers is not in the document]
SgtPooki: this is intended. see https://github.com/FilOzone/dealbot/pull/538#discussion_r3259720711

docs/checks/retrievals.md
```diff
@@ -76,12 +76,14 @@ Source: [`apps/backend/src/retrieval-addons/strategies/ipfs-block.strategy.ts`](
...
+| 3 | Content integrity via CID | Each fetched block is hash-verified against its CID during DAG traversal | Retrieval ([`retrievalStatus`](./events-and-metrics.md#retrievalStatus)) | none - if we receive non-matching bytes it's a failure | n/a (client-side) | Yes |
+| 4 | All checks pass | Check is not marked successful until all assertions pass within window | All sub-statuses above feed [`dataStorageStatus`](./events-and-metrics.md#dataStorageStatus) | n/a | [`retrievalCheckMs`](./events-and-metrics.md#retrievalCheckMs) | Yes |
+
+> `retrievalStatus` is the sub-status counter for the `/ipfs` transport stage only (assertions 2 and 3). The IPNI assertion (1) is recorded on `discoverabilityStatus`. There is no composite "retrieval check" counter; overall success comes from `dataStorageStatus`, which is `success` only when all four sub-statuses succeed. See [Deal Status Progression](./data-storage.md#deal-status-progression).
```
BigLep:
```suggestion
```

I don't think this is needed given the "Sub Status Affected" section, but if you do want to add a note about how the counters work, I would add it to https://github.com/FilOzone/dealbot/blob/main/docs/checks/retrievals.md#metrics-recorded (Metrics Recorded) section down below.

BigLep (commented):
One comment to review, but looks good to me from a docs regard. I'll leave to @silent-cipher to give the code approval.

docs/checks/retrievals.md
SgtPooki: there was some confusion from copilot here, so i think we do need to clarify a little

silent-cipher (approved):
LGTM!

PR #557, opened 2026-05-20
fix: use SP IPNI timestamps

SgtPooki:
## What changed

Dealbot now reads SP piece-status timestamps (`indexedAt` and `advertisedAt`) when providers include them, while preserving the existing local-observation fallback for older or skewed responses. Data-storage `ipniVerifyMs` now measures from the SP advertisement timestamp to filecoinpin.contact verification when that timestamp is usable, so dashboards better represent the announced-to-visible window.

The PR also expands piece-status response validation for the new SP fields, guards persisted timestamps against obvious clock skew, and updates the discoverability docs to describe the metric semantics without changing retrieval-check timing.

## How to verify

- `pnpm -C apps/backend exec tsc --noEmit --incremental false`
- `pnpm -C apps/backend exec biome ci src/deal-addons/strategies/ipni.strategy.ts src/deal-addons/strategies/ipni.types.ts src/deal-addons/strategies/ipni.strategy.spec.ts`
- `pnpm -C apps/backend exec vitest run src/deal-addons/strategies/ipni.strategy.spec.ts`

## Notes / risks

`retrievedAt` is accepted but not used for metric anchoring yet. SP clock skew beyond the guard logs `ipni_provider_timestamp_ignored` and falls back to dealbot observation time.


SgtPooki (commented):
lgtm

apps/backend/src/deal-addons/strategies/ipni.strategy.ts
```diff
@@ -673,34 +677,66 @@ export class IpniAddonStrategy implements IDealAddon<IpniMetadata> {
...
+        this.logger.warn({
+          ...dealLogContext,
+          event: "ipni_provider_timestamp_ignored",
+          message: "Ignoring invalid or skewed provider timestamp",
+          eventName,
+          providerTimestamp: timestamp,
+          uploadEndTime: uploadEndTime?.toISOString(),
+          observedAt: now.toISOString(),
+        });
+        return now;
```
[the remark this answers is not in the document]
SgtPooki: Good catch. Updated the IPNI status polling path to carry local transition observation timestamps for indexed/advertised states, then use those as the fallback when provider timestamps are missing or rejected for skew. We no longer use metrics update time as the fallback, and I added a regression test for skewed provider timestamps.

BigLep (commented):
I didn't look too closely to the code but it conceptually makes sense.

One thing that I think would be useful for debugging is: if we timeout verifying for IPNI, it would be great to have as much information in the log line about the state of the world. For example, did Curio say it was "retrieved" or not. Basically I think it would be useful to include everything in the pieceStatusReponse.

apps/backend/src/deal-addons/strategies/ipni.strategy.ts
```diff
@@ -462,40 +457,53 @@ export class IpniAddonStrategy implements IDealAddon<IpniMetadata> {
...
+          advertisedAt: providerStatus.advertisedAt ?? lastStatus.advertisedAt,
+          indexedObservedAt: lastStatus.indexedObservedAt,
+          advertisedObservedAt: lastStatus.advertisedObservedAt,
         };
 
         // Update indexedAt and advertisedAt if they have changed
-        if (currentStatus.indexed && !currentStatus.indexedAt) {
-          currentStatus.indexedAt = new Date().toISOString();
+        if (currentStatus.indexed) {
+          currentStatus.indexedObservedAt ??= observedAt;
```
juliangruber, line 507: We're mixing timestamps from this service's clock and the `pieceStatus` we fetched. Can we annotate field names with which clock the fields came from, when we mix clocks?

apps/backend/src/deal-addons/strategies/ipni.strategy.ts
```diff
@@ -673,34 +682,80 @@ export class IpniAddonStrategy implements IDealAddon<IpniMetadata> {
...
+      eventTime: Date,
+      eventName: string,
+      startTime: Date | null | undefined,
+    ): number | null => {
+      if (!startTime) return null;
+      const duration = Math.round(eventTime.getTime() - startTime.getTime());
+      const eventTimeMs = eventTime.getTime();
+      const eventTimeIso = Number.isFinite(eventTimeMs) ? eventTime.toISOString() : String(eventTime);
+
+      if (!Number.isFinite(eventTimeMs) || duration <= 0 || eventTimeMs > now.getTime() + maxFutureSkewMs) {
```
juliangruber, line 735: Multiple clocks get tricky...do you see a way how not to mix clocks at all? Or is this unavailable?

juliangruber (changes requested):

apps/backend/src/deal-addons/strategies/ipni.strategy.ts
SgtPooki: Looked at this, but going all-local kills the precision win. the new SP fields are the whole reason for this PR, and old Curio doesn't return them anyway so we can't go all-provider either. Stuck mixing, but it's only temporary. 

Once all SPs return these fields the same column stops meaning two different physical events, but we'll still be subtracting SP clock from dealbot clock (upload end is ours, index/advertise are theirs). NTP drift on healthy hosts is bounded and below dashboard granularity; the existing skew guards catch the gross cases.

see https://github.com/filecoin-project/curio/pull/1222

apps/backend/src/deal-addons/strategies/ipni.strategy.ts
SgtPooki: Good call. the mix is on purpose.. this PR partly exists to bridge the Curio rollout. Newer Curio returns indexedAt/advertisedAt from the SP clock, older Curio doesn't, and dealbot has to keep measuring both without forking. Same column ends up meaning "Curio observed" on upgraded SPs and "dealbot polled" on old ones.

I didn't want to add DB columns we'd just deprecate.. so I split PieceStatus into `Provider`/`Observed` in memory and added a source flag on the transition logs (`indexedAtSource: "provider"|"observed"` on `piece_status_indexed`/`piece_status_advertised`). Gives us BetterStack visibility into the migration without a schema we'd have to clean up.

[`70b08fc` (this PR)](https://github.com/FilOzone/dealbot/pull/557/commits/70b08fc59f27ed976f84dc05631a66d097d68deb)

juliangruber (approved):

PR #558, opened 2026-05-20
refactor: return 410 for expired pieces and remove eager cleanup

silent-cipher:
fixes #554


apps/backend/src/pull-check/pull-check.service.spec.ts
```diff
@@ -330,8 +330,8 @@ describe("PullCheckService", () => {
       // Terminal aggregate status is success.
       expect(metricsMock.recordStatus).toHaveBeenCalledWith(expect.any(Object), "success");
 
-      // Cleanup ran (forget called)
-      expect(registryMock.forget).toHaveBeenCalledWith(registration.pieceCid);
+      // Eager forget removed; pieces expire via TTL rather than being deleted at job end.
```
SgtPooki:
```suggestion
      // pieces expire via TTL rather than being deleted at job end.
```

SgtPooki (approved):
lgtm

PR #561, opened 2026-05-21
feat(retrieval): branch piece-missing on PDP pieceLive instead of always cleanup

SgtPooki:
## What changed

Retrieval pre-flight now distinguishes "SP correctly purged a piece that should be gone" from "SP failed to serve a piece it should still have."

Before this PR, a 404 from `/pdp/piece/:pieceCid/status` always marked the deal `cleaned_up`, masking real SP failures as routine cleanups.

## New flow

In `RetrievalService.performAllRetrievals`:

1. If `deal.dataSetId` or `deal.pieceId` is null, emit `failure.other` with log event `retrieval_missing_chain_ids` and bail. Backfill is a deploy prerequisite (see below).
2. Call `PDPVerifier.pieceLive(dataSetId, pieceId)`.
   - `false`: mark `cleaned_up`, emit `skipped.piece_missing`, return. No SP probe.
   - `true`: proceed.
3. SP `HEAD /pdp/piece/:pieceCid/status` (GET fallback on 405).
   - `404`: emit `failure.other`, persist a failed `Retrieval` row with `errorMessage: "SP reports piece missing but PDP pieceLive=true"` and log event `retrieval_failed_piece_missing_live`. Deal stays in the candidate pool for re-probing.
   - other: proceed to full retrieval.

`pieceLive` source: [PDPVerifier.sol:288](https://github.com/FilOzone/pdp/blob/main/src/PDPVerifier.sol#L288). Pieces scheduled for removal but not yet finalized return `true`, since the SP is still on the hook for challenges.

The documented `retrievalStatus` enum (`success | failure.timedout | failure.other | skipped.piece_missing`) is preserved. The new failure modes reuse `failure.other`; dashboards slice via log event or `errorMessage`.

## Refactor

`isDataSetLive` and its FWSS + SP-HTTP probes move out of `DealService` into a new `DatasetLivenessService`, which also owns `isPieceLive`. `DealService.isDataSetLive` is now a thin proxy.

## Backfill (completed prior to merge)

Legacy deals from before 2026-05-07 have null `piece_id` and/or `data_set_id`. Backfill was run against both environments using a one-off script (kept out of the repo). Script + run results + chain-verification spot checks: https://gist.github.com/SgtPooki/9fcf39ad2e95e022930ec3fbbbd139b1

| env | rows updated | active candidates remaining with null IDs |
|---|---|---|
| calibration | 10,774 | 0 |
| mainnet | 14,100 | 0 |

## How to verify

```bash
cd apps/backend
pnpm typecheck && pnpm vitest run && pnpm biome lint .
```

After deploy: `failure.other` on the Retrieval path should pick up SP-side piece misses that previously hid in `skipped.piece_missing`. `retrieval_missing_chain_ids` events should stay at zero.

## Notes

- `checkPieceLive` returns `true` on RPC error so a chain outage cannot cascade into bulk cleanups.
- Probe is now `HEAD` (GET fallback on 405) instead of `GET`. No body required.


SgtPooki (commented):
self review

BigLep (commented):
PR description looks good/right to me.

I do want to see "A separate backfill script must run against staging and mainnet before this PR merges. Tracking issue + script PR will be linked here."

apps/backend/src/dataset-liveness/dataset-liveness.service.ts
```diff
@@ -0,0 +1,142 @@
...
+   * Pieces that are scheduled for removal but not yet finalized still return
+   * `true` — the SP remains on the hook for challenges until the next
+   * proving period clears the scheduledRemovals queue, so it must still
+   * serve GET requests.
+   *
+   * Source: `FilOzone/pdp` PDPVerifier.sol `pieceLive`.
+   */
+  async isPieceLive(dataSetId: bigint, pieceId: bigint, signal?: AbortSignal): Promise<boolean> {
+    signal?.throwIfAborted();
+    const client = this.walletSdkService.getSynapseClient() as SynapseViemClient | null;
```
silent-cipher: Not in scope for this PR, but I think this type could potentially be moved upstream into `wallet-sdk.service.ts`.

apps/backend/src/retrieval/retrieval.service.ts
```diff
@@ -129,28 +131,87 @@ export class RetrievalService {
...
+        event: "retrieval_missing_chain_ids",
+        message: "Deal is missing dataSetId or pieceId; cannot run chain pre-check. Backfill required.",
+        dataSetId: deal.dataSetId?.toString() ?? null,
+        pieceId: deal.pieceId ?? null,
+      });
+      return [];
+    }
+
+    const pieceLive = await this.checkPieceLive(
+      BigInt(deal.dataSetId.toString()),
```
silent-cipher: Is this necessary given that `dataSetId` already uses `bigintTransformer`?

silent-cipher (approved):
Looks good to me!

apps/backend/src/dataset-liveness/dataset-liveness.service.ts
silent-cipher: #564

apps/backend/src/retrieval/retrieval.service.ts
silent-cipher: #564

PR #564, opened 2026-05-22
fix(retrieval): filter random deal selection by wallet address

silent-cipher:
https://filecoinproject.slack.com/archives/C07CGTXHHT4/p1779381027144089

This PR fixes the issue discussed in the above Slack thread.
```
Dealbot’s payer wallet was rotated on 2026-03-29 (calibration) and 2026-04-02 (mainnet). Deals created by the prior wallet were left in dealbot’s database without a cleaned_up marker, so the retrieval scheduler kept probing them. Most of those datasets had been terminated on chain after the rotation, and the SPs had correctly purged the pieces. Each probe returned 404 piece missing, dragging the displayed success rate down.
```

This change filters deals before retrieval scheduling to only consider deals associated with the currently active Dealbot wallet.


SgtPooki (approved):
nice

PR #567, opened 2026-05-22
fix(deal): wire abort signal and graceful pg-boss drain

SgtPooki:
## What changed

Two related fixes that close the gap between in-flight deal jobs and graceful worker shutdown.

**Abort signal reaches `executeUpload`.** The deal-creation path called `executeUpload` without passing the per-job `AbortController` signal, so the 360s `dealJobTimeoutSeconds` could fire without interrupting a stalled upload. The call now mirrors the dataset-creation path: wrapped in `awaitWithAbort` and `signal` passed inside `UploadExecutionOptions`. A hung upload now rejects within the configured job timeout.

**Double `dataStorageStatus` emit guard.** If `postProcessDeal` threw after the success metric was recorded, the outer catch re-emitted `failure.*` for the same deal, inflating failure counts. A `dataStorageStatusEmitted` flag (matching the existing `retrievalStatusEmitted` pattern) prevents the duplicate emit.

**Graceful pg-boss drain.** `boss.stop()` was called with no options, which capped drain at the pg-boss default of 30s, far shorter than `dealJobTimeoutSeconds`. The shutdown handler now:

- Passes `{ graceful: true, timeout: longestJobTimeoutSec + 60s }` so handlers can reach terminal state.
- Sleeps for `SHUTDOWN_FINAL_SCRAPE_DELAY_SECONDS` (default 35s) after drain, letting Prometheus capture the terminal counter increments before the process exits.

Pairs with FilOzone/infra#208, which bumps `terminationGracePeriodSeconds` and adds `publishNotReadyAddresses: true` on the worker Service.

## How to verify

```bash
pnpm --filter dealbot-backend test
pnpm --filter dealbot-backend exec tsc --noEmit
pnpm exec biome check apps/backend/src/deal/deal.service.ts apps/backend/src/jobs/jobs.service.ts
```

After deploying alongside the infra PR, roll the worker Deployment in staging and confirm:

- No deals left in `uploaded` / `piece_added` / `piece_confirmed` after rollout completes.
- `dataStorageStatus{value=pending}` counter increments are matched by `success` or `failure.*` over the rollout window.

## Notes / risks

- The doc for `SHUTDOWN_FINAL_SCRAPE_DELAY_SECONDS` calls out the relationship with `terminationGracePeriodSeconds`. Operators tuning either need to keep them in sync.
- Drain extends pod termination from ~30s to up to ~480s. Argo / k8s rollout dashboards will show longer terminating states; this is expected.

### DIFF
--- a/apps/backend/src/config/app.config.ts
+++ b/apps/backend/src/config/app.config.ts
@@ -96,7 +96,7 @@ export const configValidationSchema = Joi.object({
   // Seconds to hold the process alive after pg-boss drain completes, so Prometheus
   // captures at least one scrape of the terminal counter increments emitted during
   // shutdown. Default 35 covers the 30s ServiceMonitor interval plus a 5s buffer.
-  SHUTDOWN_FINAL_SCRAPE_DELAY_SECONDS: Joi.number().min(0).default(35),
+  SHUTDOWN_FINAL_SCRAPE_DELAY_SECONDS: Joi.number().min(0).max(300).default(35),
   IPFS_BLOCK_FETCH_CONCURRENCY: Joi.number().integer().min(1).max(32).default(6),
 
   // Pull Check
--- a/apps/backend/src/config/app.config.ts
+++ b/apps/backend/src/config/app.config.ts
@@ -93,6 +93,10 @@ export const configValidationSchema = Joi.object({
   DEAL_JOB_TIMEOUT_SECONDS: Joi.number().min(120).default(360), // 6 minutes max runtime for data storage jobs (TODO: reduce default to 3 minutes)
   RETRIEVAL_JOB_TIMEOUT_SECONDS: Joi.number().min(60).default(60), // 1 minute max runtime for retrieval jobs (TODO: reduce default to 30 seconds)
   DATA_SET_CREATION_JOB_TIMEOUT_SECONDS: Joi.number().min(60).default(300), // 5 minutes max runtime for dataset creation jobs
+  // Seconds to hold the process alive after pg-boss drain completes, so Prometheus
+  // captures at least one scrape of the terminal counter increments emitted during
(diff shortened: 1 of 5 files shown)


apps/backend/src/config/app.config.ts
```diff
@@ -93,6 +93,10 @@ export const configValidationSchema = Joi.object({
   DEAL_JOB_TIMEOUT_SECONDS: Joi.number().min(120).default(360), // 6 minutes max runtime for data storage jobs (TODO: reduce default to 3 minutes)
   RETRIEVAL_JOB_TIMEOUT_SECONDS: Joi.number().min(60).default(60), // 1 minute max runtime for retrieval jobs (TODO: reduce default to 30 seconds)
   DATA_SET_CREATION_JOB_TIMEOUT_SECONDS: Joi.number().min(60).default(300), // 5 minutes max runtime for dataset creation jobs
+  // Seconds to hold the process alive after pg-boss drain completes, so Prometheus
+  // captures at least one scrape of the terminal counter increments emitted during
+  // shutdown. Default 35 covers the 30s ServiceMonitor interval plus a 5s buffer.
+  SHUTDOWN_FINAL_SCRAPE_DELAY_SECONDS: Joi.number().min(0).default(35),
```
silent-cipher:
```suggestion
  SHUTDOWN_FINAL_SCRAPE_DELAY_SECONDS: Joi.number().min(0).max(300).default(35),
```

apps/backend/src/deal/deal.service.ts
silent-cipher, line 375: maybe signal abort check after `isDataSetLive` contract call?

silent-cipher (approved):
LGTM!

apps/backend/src/deal/deal.service.ts
SgtPooki: the next call is awaitWithAbort that will throw immediately on an aborted signal, so i'll leave for now

Issue #484, opened 2026-04-27
Redesign ipniVerifyMs to disambiguate verify outcomes and tighten histogram buckets


SgtPooki:
## Problem

`ipniVerifyMs` is a single histogram with no outcome dimension. It records duration regardless of result (verified, signal-aborted, monitoring-skipped, error). Dashboards cannot filter to success-only or compute success-rate from it. Bucket boundaries `[10, 50, 100, 500, 1000, 2000, 5000, 10000, 30000, 60000, 120000, 300000]` jump from 60s → 120s → 300s, so `quantileInterpolatedWeighted` linearly extrapolates a deep tail (e.g. p90 ≈ 264s) from a small number of >120s samples.

Two related call paths feed the metric:

1. **Verify path** (`apps/backend/src/ipni/ipni-verification.service.ts:25-142`): polls IPNI indexer, bounded by `IPNI_VERIFICATION_TIMEOUT_MS=60000` via `AbortSignal.timeout`. Observed at `apps/backend/src/deal-addons/strategies/ipni.strategy.ts:406` and `apps/backend/src/retrieval/retrieval.service.ts:427`.
2. **Monitoring fallback** (`apps/backend/src/deal-addons/strategies/ipni.strategy.ts:303-332`): when `monitorPieceStatus` (separate `POLLING_TIMEOUT_MS=120000`) times out, the verify call still runs after — but the wall-time framing differs and the metric cannot distinguish.

### A. Add `outcome` label to `ipniVerifyMs`

`outcome ∈ { verified, timeout, error }` — observed for every attempted verification. Lets dashboards filter `{outcome=\"verified\"}` for true latency SLI. Recording timeout durations under `outcome=timeout` exposes potential AbortSignal-leak bugs (durations exceeding the budget by a wide margin).

### B. Add `ipniVerifySkippedTotal` counter

Incremented when monitoring failure prevents a verify attempt. Avoids polluting the histogram with synthetic-zero or \"skipped\" durations.

### C. Tighten histogram buckets

Current: `[10, 50, 100, 500, 1000, 2000, 5000, 10000, 30000, 60000, 120000, 300000]`
Proposed: `[10, 50, 100, 500, 1000, 2000, 5000, 10000, 30000, 60000, 90000, 120000, 180000, 240000, 300000]`

Closes the 60s→120s→300s gap that drives misleading p90 interpolation.

### D. (Optional) Separate `ipniMonitorStatusMs` histogram

[...]

PR #568, opened 2026-05-22
refactor(metrics): rename outcome label to value

SgtPooki:
### What changed

Rename the `ipniVerifyMs` histogram label `outcome` to `value` to match the convention used by all other check metrics in `docs/checks/events-and-metrics.md`. Label values (`verified` / `timeout` / `error`) unchanged. The internal `IpniVerifyOutcome` type name is kept.

### How to verify

- `pnpm typecheck`
- `pnpm test`
- Confirm `ipniVerifyMs{value="verified"}` appears in Prometheus output after deploy; no `outcome` label remains.

### Notes / risks

No production dashboards reference `ipniVerifyMs{outcome=...}` yet (dashboard work is the open scope on #484), so the rename is safe to land before chart authoring.

Refs #484.


SgtPooki (commented):
self review

PR #569, opened 2026-05-23
feat: add network label to prometheus metrics

silent-cipher:
Built on top of #463

### DIFF
--- a/apps/backend/src/metrics-prometheus/check-metric-labels.ts
+++ b/apps/backend/src/metrics-prometheus/check-metric-labels.ts
@@ -1,3 +1,5 @@
+import { Network } from "../common/types.js";
+
 export type CheckType = "dataStorage" | "retrieval" | "dataRetention" | "dataSetCreation" | "pullCheck";
 export type ProviderStatus = "approved" | "unapproved";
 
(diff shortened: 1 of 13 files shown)


apps/backend/src/metrics-prometheus/check-metric-labels.ts
```diff
@@ -1,3 +1,5 @@
+import { Network } from "../common/types.js";
+
```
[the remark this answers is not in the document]
silent-cipher: I've added the manual `type` import here. Just a quick note that this wasn't a blocker. The tests and build were perfectly green. As a reminder, we intentionally disabled the `useImportType` rule in biome because NestJS relies on standard class imports for DI.

apps/backend/src/pull-check/pull-check.service.ts
```diff
@@ -68,6 +68,7 @@ export class PullCheckService {
     const providerInfo = this.validateProviderInfo(spAddress);
     const labels = buildCheckMetricLabels({
       checkType: "pullCheck",
+      network: this.configService.get("blockchain.network", { infer: true }),
```
SgtPooki: nit: inconsistent with other (more majority) usage with `get("blockchain", {infer:true}).network`

```suggestion
      network: this.configService.get("blockchain", {infer:true}).network,
```

SgtPooki (approved):

apps/backend/src/pull-check/pull-check.service.ts
silent-cipher: [fd4713a](https://github.com/FilOzone/dealbot/pull/569/commits/fd4713af17cfde544975ee2c122068795a14c2df)

PR #576, opened 2026-05-26
feat(web): add multi-network support with network switcher

silent-cipher:
Built on top of #570 

**Note**- Need multi network support propagated to betterstack dashboard.

### DIFF
--- a/apps/web/src/components/shared/Network/NetworkSwitcher.test.tsx
+++ b/apps/web/src/components/shared/Network/NetworkSwitcher.test.tsx
@@ -1,146 +1,50 @@
-import { server } from "@test/mocks/server";
-import { render, screen, waitFor } from "@testing-library/react";
-import { HttpResponse, http } from "msw";
-import { MemoryRouter } from "react-router-dom";
-import { describe, expect, it } from "vitest";
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { describe, expect, it, vi } from "vitest";
 import NetworkSwitcher from "./NetworkSwitcher";
 
-const configUrl = "/api/config";
-
 describe("NetworkSwitcher", () => {
-  it("shows current network and a link to switch to the other deployment (mainnet → calibration)", async () => {
-    server.use(
-      http.get(configUrl, () =>
-        HttpResponse.json({
-          network: "mainnet",
-          jobs: {},
-        }),
-      ),
-    );
-
-    render(
-      <MemoryRouter>
-        <NetworkSwitcher />
-      </MemoryRouter>,
-    );
-
-    const switchLink = await screen.findByRole("link", { name: /Switch to Calibration/i });
-    expect(switchLink).toHaveAttribute("href", "https://staging.dealbot.filoz.org");
+  it("renders nothing when only one network is active", () => {
+    const { container } = render(<NetworkSwitcher networks={["mainnet"]} selected="mainnet" onChange={vi.fn()} />);
+    expect(container.firstChild).toBeNull();
   });
 
-  it("offers switching to mainnet when current instance monitors calibration", async () => {
-    server.use(
-      http.get(configUrl, () =>
-        HttpResponse.json({
-          network: "calibration",
-          jobs: {},
-        }),
-      ),
-    );
-
-    render(
-      <MemoryRouter>
-        <NetworkSwitcher />
-      </MemoryRouter>,
-    );
-
-    const switchLink = await screen.findByRole("link", { name: /Switch to Mainnet/i });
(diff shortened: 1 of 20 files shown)


apps/web/src/pages/Landing.tsx
```diff
@@ -80,16 +81,22 @@ const getConfig = (network: Network | null) => {
...
   const {
     dashboardUrl,
     dashboardUrlInvalid,
     approvedSpDashboardUrl,
     approvedSpDashboardUrlInvalid,
     logsUrl,
     logsUrlInvalid,
-  } = getConfig(network);
-  const { providers: providersResponse, loading: providersLoading, error: providersError } = useProvidersList(0, 500);
+  } = getConfig(selectedNetwork);
```
silent-cipher, line 93: This seems to be a real problem - how do we pick the dashboard url? The metrics/logs url is per environment, which is correct. But each environment can have two networks now, so same provider ID on calibration and mainnet are not the same provider. They're different. And, better stack dashboard doesn't know about this distinction yet.

apps/web/src/pages/Landing.tsx
SgtPooki: only thing that should really be fixed before merge:

we should add a `const providersPending = configLoading || selectedNetwork === null || providersLoading;` above and then do

```suggestion
{providersError && <p className="text-sm text-destructive">{providersError}</p>}
{(providersLoading || providersPending) && <p className="text-sm text-muted-foreground">Loading providers…</p>}
{!providersLoading &&
```

or just put that logic in `providersLoading`

apps/web/src/pages/Landing.tsx
SgtPooki: Our calib and mainnet dashboards are separate IDs, not one dashboard you can `vs[network]=` filter, so it has to be distinct URLs per network. Just finish the approvedSpSources pattern in getConfig(), and keep a fallback to the bare global so live deployments don't go dark before infra sets the new vars:

`approvedSpSources[network].runtime ?? runtimeConfig?.DASHBOARD_URL`

Touches Landing.tsx getConfig(), vite-env.d.ts, docker-entrypoint.sh, .env.example / kustomize local, plus the infra env.

Fine as a fast-follow if you want to keep this PR on the switcher UI. Global URLs are correct for single-network deploys, the gap only bites at 2 networks.

silent-cipher: Created a new PR for per-network BetterStack dashboard and logs url configuration - #625

apps/web/src/pages/Landing.tsx
silent-cipher: [2b359c2](https://github.com/FilOzone/dealbot/pull/576/commits/2b359c2017cdc4caaf055403232e2a2d28cb01e9)

apps/web/src/pages/Landing.tsx
```diff
@@ -80,16 +81,24 @@ const getConfig = (network: Network | null) => {
 };
 
 export default function Landing() {
-  const { network } = useNetworkConfig();
+  const { activeNetworks, loading: configLoading } = useActiveNetworks();
+  const [selectedNetwork, setSelectedNetwork] = useSelectedNetwork(activeNetworks);
```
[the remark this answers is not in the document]
silent-cipher: [6ef6d83](https://github.com/FilOzone/dealbot/pull/576/commits/6ef6d838feb7ad6f85c9eb5d96daee4720116e60)

apps/web/src/types/providers.ts
```diff
@@ -28,6 +28,7 @@ export interface PDPOffering {
  */
 export interface Provider {
   address: string;
+  network: string;
```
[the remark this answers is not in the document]
silent-cipher: [6ef6d83](https://github.com/FilOzone/dealbot/pull/576/commits/6ef6d838feb7ad6f85c9eb5d96daee4720116e60)

apps/web/src/components/shared/Network/NetworkSwitcher.test.tsx
```diff
@@ -1,131 +1,50 @@
...
+    const onChange = vi.fn();
+    render(<NetworkSwitcher networks={["mainnet", "calibration"]} selected="mainnet" onChange={onChange} />);
 
-    const { container } = renderSwitcher();
-
-    await waitFor(() => {
-      expect(container.firstChild).toBeNull();
-    });
+    await userEvent.click(screen.getByRole("tab", { name: /calibration/i }));
+    expect(onChange).toHaveBeenCalledWith("calibration");
```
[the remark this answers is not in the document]
silent-cipher: [6ef6d83](https://github.com/FilOzone/dealbot/pull/576/commits/6ef6d838feb7ad6f85c9eb5d96daee4720116e60)

PR #487, opened 2026-04-28
feat: anon piece selection and retrieval

dennis-tra:
Hi folks,

this PR adds the anon retrieval flow from https://github.com/FilOzone/dealbot/issues/427. It is a follow-up of:

- https://github.com/FilOzone/dealbot/pull/469 - added the subgraph)
- https://github.com/FilOzone/dealbot/pull/459 - the original large PR which contained the subgraph and anon retrieval flow (I've closed that PR).

The main logic is in `./apps/backend/retrieval-anon`:

- `anon-retrieval.service.ts` - "called" on a schedule as a job. It starts the retrieval process (select piece, fetch piece, validate car, store result)
- `anon-piece-selector.service.ts` - implements the subgraph query logic
- `car-validation.service.ts` - parses the piece bytes as a CAR, checks IPNI availability, fetches `k` blocks from that car and validates their hashes
- `piece-retreival.service.ts` - implements the HTTP request to download the piece and CommP validation

The anonymous piece selection logic works as follows:

The `retrievalAnon` check probes an SP for non-dealbot pieces so we can detect SPs that behave well even if the teacher is not watching. To do this fairly, the piece selection should satisfy the following requirements:

1. uniform random across the SP's entire active pieces (not biased toward recent writes, specific payers, or specific sizes).
2. Prefer `withIPFSIndexing` pieces (so CAR/IPNI validation has something to check) but still exercise non-indexed pieces so an SP can't optimise only its CAR corpus.
3. Cover a realistic spread of piece sizes: big enough for useful bandwidth measurements, not so big that SPs with only small deals are skipped.
5. Avoid immediately re-testing the same piece across consecutive checks.

How it works in practice:

Every Root entity in the subgraph carries a `sampleKey = keccak256(setId-rootId)` populated once at insert time. Because `keccak256` is uniform over 256 bits and independent of creation order/size/dataset,
`sampleKey` sorts roots into a uniform random permutation that is stable across queries.

This is necessary because you cannot just select a random element from a range query in GraphQL. If we knew the total number of pieces we could define a random `skip` value but this is also capped at 5000. I've read that it becomes very inefficient at higher values. This would also require a non-trivial bookkeeping of active pieces/datasets counts. The `sampleKey` is much easier.

Drawing a sample looks like this:

1. Pick a size bucket (small < 20 MiB, medium 20 MiB to 100 GiB, large 100 MiB to 500 MiB) by weighted random — weights 20% / 50% / 30% respectively.
[...]

PR #582, opened 2026-05-29
feat(subgraph): track payment-active state for cheaper anon-retrieval queries

dennis-tra:
The subgraph part of #579
The backend part is here https://github.com/probe-lab/dealbot/pull/12 and depends on https://github.com/FilOzone/dealbot/pull/487 to land first.

## Why

In the pending retrieval anon implementation why draw candidate pieces by running multiple attempts and then hoping that we draw a pieces whose payment hasn't terminated yet. The reason for this is that we'd need to pass the current epoch to the query to filter the result set server side which isn't available. Secondly, we'll get served the current epoch together with the query result but cannot make use of that field during the querying process so we can only apply the filter condition after we got served the data.

## What

In this PR we add a callback to the subgraph code that fires on every block (`handleBlock`) which "cleans up" all datasets that have expired at that epoch by setting a new `isPaymentActive` flag to false. The rest of the code is just bookkeeping of that flag as a response to various events.

How it is maintained:
- `PDPPaymentTerminated` with future endEpoch schedules the dataset into a `PendingPaymentTermination` bucket keyed by epoch; if endEpoch is already past, the flag flips immediately.
- A new `handleBlock` attached to the FWSS data source loads the bucket for the current block (O(1) load returning null in the steady state) and flips matching datasets to `isPaymentActive: false`.
- `ServiceTerminated` also flips the flag and unschedules from the bucket.
- Re-extension (a new PDPPaymentTerminated after a prior epoch elapsed) sets the flag back to true and reschedules.

Backwards compatible: `pdpPaymentEndEpoch` stays on the schema unchanged and is still populated. Existing clients comparing it to the current epoch keep working; new clients can use the cheaper `isPaymentActive` filter instead.

## Deployment

Deployed new version of this subgraph to:

- `calibration`: https://api.goldsky.com/api/public/project_cmppqidloyxtu01tgc9b3e3w4/subgraphs/dealbot-calibration/0.1.0/gn
- `mainnet`: https://api.goldsky.com/api/public/project_cmppqidloyxtu01tgc9b3e3w4/subgraphs/dealbot-mainnet/0.1.0/gn

I've also reset the version to `0.1.0` because I've deployed them into a new Goldsky account which I'll plan to hand over to FilOZ.

### DIFF
--- a/apps/subgraph/src/fwss.ts
+++ b/apps/subgraph/src/fwss.ts
@@ -1,4 +1,4 @@
-import { BigInt, log } from "@graphprotocol/graph-ts";
+import { BigInt, Bytes, ethereum, log, store } from "@graphprotocol/graph-ts";
 import {
   DataSetCreated as DataSetCreatedEvent,
   DataSetServiceProviderChanged as DataSetServiceProviderChangedEvent,
@@ -6,8 +6,14 @@ import {
   PieceAdded as PieceAddedEvent,
   ServiceTerminated as ServiceTerminatedEvent,
 } from "../generated/FilecoinWarmStorageService/FilecoinWarmStorageService";
-import { DataSet, Root } from "../generated/schema";
-import { arrayContains, extractMetadataValue, getProofSetEntityId, getRootEntityId } from "./helpers";
+import { DataSet, PendingPaymentTermination, Root } from "../generated/schema";
+import {
+  arrayContains,
+  extractMetadataValue,
+  getProofSetEntityId,
+  getRootEntityId,
+  getBucketId,
+} from "./helpers";
 import { DataSetStatus } from "./types";
 
 // ---- Handlers -------------------------------------------------------------
@@ -28,6 +34,7 @@ export function handleFwssDataSetCreated(event: DataSetCreatedEvent): void {
     // PDPVerifier-level non-null defaults; handleDataSetCreated will overwrite.
     ds.owner = event.params.serviceProvider;
     ds.isActive = true;
+    ds.isPaymentActive = true;
     ds.status = DataSetStatus.EMPTY;
     ds.nextDeadline = BigInt.zero();
     ds.maxProvingPeriod = BigInt.zero();
@@ -63,6 +70,12 @@ export function handleFwssServiceTerminated(event: ServiceTerminatedEvent): void
   }
 
   ds.isActive = false;
+  ds.isPaymentActive = false;
+  const scheduledEpoch = ds.pdpPaymentEndEpoch;
+  if (scheduledEpoch !== null) {
+    unschedulePaymentTermination(ds.id, scheduledEpoch);
+  }
+
   ds.save();
 }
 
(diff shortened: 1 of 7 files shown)


apps/subgraph/src/fwss.ts
```diff
@@ -63,6 +70,12 @@ export function handleFwssServiceTerminated(event: ServiceTerminatedEvent): void
   }
 
   ds.isActive = false;
+  ds.isPaymentActive = false;
+  const scheduledEpoch = ds.pdpPaymentEndEpoch;
+  if (scheduledEpoch !== null) {
+    unschedulePaymentTermination(ds.id, scheduledEpoch);
+  }
```
silent-cipher, line 77: Wouldn't `ServiceTerminated` always co-fires with `PDPPaymentTerminated` in the same transaction? This is because the existing condition in fwss that a rail can only be terminated by FWSS contract (https://github.com/FilOzone/filecoin-services/blob/8bed7215738994acd69489930fe6cd3dad866d5e/service_contracts/src/FilecoinWarmStorageService.sol#L1587).
So, this handler will always flip `isPaymentActive` to false at the termination block and removes the bucket before `handleBlock` could ever reach `endEpoch`.

apps/subgraph/src/fwss.ts
```diff
@@ -73,7 +86,27 @@ export function handleFwssPdpPaymentTerminated(event: PDPPaymentTerminatedEvent)
     return;
   }
 
+  const previousEpoch = ds.pdpPaymentEndEpoch;
+  if (previousEpoch !== null && previousEpoch.notEqual(event.params.endEpoch)) {
+    unschedulePaymentTermination(ds.id, previousEpoch);
+  }
```
silent-cipher, line 92: This looks like a dead defensive complexity to me. The FWSS contract guards rail's re-termination. 
https://github.com/FilOzone/filecoin-services/blob/8bed7215738994acd69489930fe6cd3dad866d5e/service_contracts/src/FilecoinWarmStorageService.sol#L1018
https://github.com/FilOzone/filecoin-services/blob/8bed7215738994acd69489930fe6cd3dad866d5e/service_contracts/src/FilecoinWarmStorageService.sol#L1594

apps/subgraph/src/fwss.ts
```diff
@@ -89,3 +122,99 @@ export function handleFwssDataSetServiceProviderChanged(event: DataSetServicePro
...
   ds.save();
 }
+
+/**
+ * Block handler: flip `isPaymentActive` to false on every DataSet whose
+ * `pdpPaymentEndEpoch` equals the current block. In the steady state this
+ * is a single bucket load returning null, so the per-block cost is O(1).
+ */
+export function handleBlock(block: ethereum.Block): void {
+  const bucketId = getBucketId(block.number);
```
silent-cipher, line 132: This handler uses exact equality (`getBucketId(block.number)` and `pdpPaymentEndEpoch.notEqual(block.number)`) to load and flip `isPaymentActive`. But, Filecoin has null rounds (no tipset at a given epoch height), so if endEpoch lands on a skipped height, `handleBlock` is never invoked with `block.number == endEpoch`, the bucket is never consumed, and the dataset stays `isPaymentActive: true` forever.

PR #571, opened 2026-05-25
chore: release to production (main)

FilOzzy:
:robot: I have created a release *beep* *boop*
---

<details><summary>backend: 1.8.0</summary>

### Features

* add cid.contact ipni verification after filecoinpin.contact check ([#580](https://github.com/FilOzone/dealbot/issues/580)) ([9ca4ffa](https://github.com/FilOzone/dealbot/commit/9ca4ffa2e7027097a541719fb8e9ff7b6861733f))
* add pull_checks table to clickhouse ([#522](https://github.com/FilOzone/dealbot/issues/522)) ([4c1a9ca](https://github.com/FilOzone/dealbot/commit/4c1a9cad05cda9bbd675f3e2d72ce3b9cf08c91b))
* store deal retrieval metrics in clickhouse ([#595](https://github.com/FilOzone/dealbot/issues/595)) ([be18abe](https://github.com/FilOzone/dealbot/commit/be18abedaa0d26fb64c6b1cca43acfbb44917cf3))

### Bug Fixes

* jobs report success during dependency outages ([#592](https://github.com/FilOzone/dealbot/issues/592)) ([d934ba6](https://github.com/FilOzone/dealbot/commit/d934ba690da41833ac68aaab293efb1bf1da64d2))
* use SP IPNI timestamps ([#557](https://github.com/FilOzone/dealbot/issues/557)) ([c621f7f](https://github.com/FilOzone/dealbot/commit/c621f7fb2114b5420c25695fc494be00b086de8f))
</details>

<details><summary>web: 1.3.1</summary>

### Miscellaneous

* **web:** migrate to SWR for data fetching ([#585](https://github.com/FilOzone/dealbot/issues/585)) ([33f62be](https://github.com/FilOzone/dealbot/commit/33f62bef521fdb95f34799fdda6dc94f5286cb69))
</details>

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).

PR #588, opened 2026-06-01
feat: data set lifecycle job

silent-cipher:
## Summary

Adds a new `data_set_lifecycle_check` job that exercises the full `createDataSet` → `terminateService` lifecycle in a single self-contained run. It creates a throwaway empty data set, terminates it, and waits for transaction receipt. This runs regardless of how many datasets a provider already has.

The job is calibration-only by default (`DATASET_LIFECYCLE_CHECK_ENABLED=false` on mainnet).

## Changes

- `jobs.service.ts` — new `data_set_lifecycle_check` job type: scheduling, worker registration, singleton-per-SP enforcement
- `data-set-lifecycle.service.ts` — `runLifecycleCheck`: creates a tagged empty data set (`dealbotLifecycleCheck` metadata key), calls `terminateService`, and waits for the transaction receipt
- `check-metrics.service.ts` — new `dataSetLifecycleCheckStatus` and `dataSetLifecycleCheckMs` metrics
-  `app.config.ts` / `.env.example` — three new config variables (see below)
- Docs — `docs/checks/data-set-lifecycle-check.md`, updated `environment-variables.md`, `jobs.md`, `checks/README.md`, `events-and-metrics.md`

## New Config Variables

- `DATASET_LIFECYCLE_CHECK_ENABLED` – default: `true` (calibration) / `false` (mainnet) , enable/disable the job
- `DATASET_LIFECYCLE_CHECKS_PER_SP_PER_HOUR` – check rate per provider
- `DATA_SET_LIFECYCLE_CHECK_JOB_TIMEOUT_SECONDS` – default: 600 secs, max job runtime before forced abort

Closes #586


docs/data-set-termination.md
```diff
@@ -0,0 +1,216 @@
...
+`checkType`, `providerId`, `providerName`, `providerStatus`.
+
+For termination metrics, `checkType=dataSetTermination`. For creation metrics referenced below, `checkType=dataSetCreation`.
+
+#### Creation metrics (primary signal)
+
+| Metric | `value` labels | What to watch for |
+|--------|---------------|-------------------|
+| `dataSetCreationStatus`| `pending`, `success`, `failure.timedout`, `failure.other`, `skipped.no_candidate` | `success` count should rise in the interval after each termination; persistent `failure.*` after a termination indicates a `createDataSet` regression |
+| `dataSetCreationMs`| — | Latency histogram for `createDataSetWithPiece`; spikes after termination may indicate on-chain congestion |
```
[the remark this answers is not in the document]
silent-cipher: [c537509](https://github.com/FilOzone/dealbot/pull/588/commits/c5375095d78b0067638b580d337ceb0ae6a3dcf9)

docs/data-set-termination.md
```diff
@@ -0,0 +1,220 @@
+# Data Set Termination Job
+
+This doc proposes a calibration-only `data_set_termination` job that periodically terminates a dealbot managed dataset so the existing `data_set_creation` job naturally recreates it. The goal is to keep dealbot continuously exercising the on-chain `createDataSet` lifecycle instead of only creating datasets until a steady-state cap is reached.
```
BigLep: Hyperlink to `data_set_creation` docs

docs/data-set-termination.md
```diff
@@ -0,0 +1,220 @@
...
+
+> **Note**: this design does **not** attempt `PDPVerifier.deleteDataSet`, which is SP-initiated; if `deleteDataSet` canary coverage is required for #586, that would need a different approach.
```
BigLep:
```suggestion
> **Note**: this design does **not** attempt `PDPVerifier.deleteDataSet`, which is SP-initiated. 
```

This is a good callout, but I think we're agreeing here not do `deleteDataSet` so we don't need to say more.

docs/data-set-termination.md
```diff
@@ -0,0 +1,220 @@
...
+
+## Summary
+
+- `data_set_termination` is a calibration-only job that periodically terminates one managed dataset slot per provider.
```
BigLep: I think it should be calibration-only by default, but it should be possible to run in mainnet if someone enables it.

docs/data-set-termination.md
```diff
@@ -0,0 +1,220 @@
...
+- Together with [`data_set_creation`](./data-set-creation.md), the two jobs form a bounded loop that keeps the `createDataSet` on-chain path continuously exercised as a canary.
+- The job terminates **at most one dataset per invocation**; `data_set_creation` handles replenishment on its next scheduled tick.
+- It runs on the shared `sp.work` queue with `singletonKey=spAddress`, so it cannot race with `deal`, `retrieval`, `piece_cleanup`, `pull_check`, or `data_set_creation` for the same provider.
```
BigLep: Rather than list them all out (which is going to get stale as we add jobs), maybe just say 


```suggestion
- It runs on the shared `sp.work` queue with `singletonKey=spAddress`, so it cannot race with other jobs for the same provider.
```

docs/data-set-termination.md
```diff
@@ -0,0 +1,220 @@
...
+- Schedule creation is gated by `NETWORK=calibration` and a non-empty canary window (`MIN_NUM_DATASETS_FOR_CHECKS - DATA_SET_TERMINATION_MIN_INDEX > 0`).
+- Slots below `DATA_SET_TERMINATION_MIN_INDEX` are never touched, keeping a stable baseline for ongoing checks.
+
+## Problem Context
+
+During the PDPVerifier v3.4.0 rollout, `createDataSet` broke on calibration and mainnet. Dealbot did not detect the calibration outage because providers had already reached the steady-state cap of managed datasets ([`MIN_NUM_DATASETS_FOR_CHECKS`](./environment-variables.md#min_num_datasets_for_checks)). Once the cap was reached, `data_set_creation` stopped exercising the on-chain create path, so the canary value of that job disappeared.
```
BigLep:
```suggestion
During the PDPVerifier v3.4.0 rollout, `createDataSet` broke on calibration and mainnet. Dealbot did not detect the calibration outage because providers had already reached the steady-state cap of managed datasets ([`MIN_NUM_DATASETS_FOR_CHECKS`](./environment-variables.md#min_num_datasets_for_checks)). Once the cap was reached, `data_set_creation` stopped exercising the on-chain create path, so the canary value of that job disappeared. (See [post mortem](https://app.notion.com/p/filecoindev/2026-05-28-Hotfix-FWSS-v1-2-0-compatibility-with-PDPVerifier-v3-4-0-901dc41950c1820d99c88157214dec5d) for more info.)
```

docs/data-set-termination.md
```diff
@@ -0,0 +1,220 @@
...
+
+The missing capability is not more creation logic. The missing capability is a controlled way to create fresh demand for creation again.
+
+## Goals
+
+- Continuously exercise the calibration `createDataSet → terminateService → createDataSet` lifecycle.
+- Reuse the existing `data_set_creation` job as the replenishment mechanism.
+- Minimize disruption to ongoing deal and retrieval checks.
+- Make termination cadence explicitly configurable so the expected create cadence can be reasoned about.
+- Ensure the job cannot run on mainnet.
```
BigLep:
```suggestion
- Ensure the job doesn't run on mainnet by default.
```

BigLep (commented):
Note: I submitted this prematurely. I am still reviewing and will send another review when I'm done reading through.

I don't think we should say that this closes https://github.com/FilOzone/dealbot/issues/586

We still need to do the implementation work and we should make sure we have visibility on this job on the internal dealbot dashboard.

docs/data-set-termination.md
```diff
@@ -0,0 +1,220 @@
...
+
+- `DATA_SET_TERMINATION_JOB_TIMEOUT_SECONDS`
+  - max runtime for one termination job invocation
+
+- `DATA_SET_TERMINATION_MIN_INDEX`
+  - the lowest slot index eligible for termination (inclusive)
+  - default: `1` — only the baseline slot (index `0`) is protected
+  - slots `0` through `DATA_SET_TERMINATION_MIN_INDEX - 1` are never touched by this job
+  - example: `MIN_NUM_DATASETS_FOR_CHECKS = 10`, `DATA_SET_TERMINATION_MIN_INDEX = 5` → slots 0–4 are stable, slots 5–9 cycle as the canary window
+  - set to `MIN_NUM_DATASETS_FOR_CHECKS` to disable termination entirely — the canary window becomes empty and no schedule is created
```
BigLep: I guess this works, but maybe we use a more obvious "this is disabled" value like "-1" or even a string "DISABLED"?

docs/data-set-termination.md
```diff
@@ -0,0 +1,220 @@
...
+The scheduling model mirrors `data_set_creation`:
+
+- queue: shared `sp.work`
+- `singletonKey=spAddress`
+
+Sharing the singleton with other SP jobs prevents termination from racing with a `deal`, `retrieval`, `pull_check`, `piece_cleanup`, or `data_set_creation` job for the same provider.
+
+The schedule is only upserted when all of the following are true:
+
+- `NETWORK=calibration`
```
BigLep: I don't know if we want to enforce this in code.

Other ideas:
- if the network is mainnet, maybe we log an info/warning (e.g., "Dataset deletion is enabled. Mainnet funds will likely drain faster as a result of ensuing repeated createDataSet fees.") or
- if the network is mainnet, require a more explicit opt in: (e.g., ENABLE_DATASET_DELETION=TRUE") which by default is false for mainnet. Basically I don't want our code to get in the way if we or others decide they want to run deletion jobs against mainnet.

docs/data-set-termination.md
```diff
@@ -0,0 +1,220 @@
...
+
+### Proposed handler algorithm
+
+For one provider, one invocation of `data_set_termination` works like this:
+
+1. Check that the network is calibration. If not, log skip and exit.
+2. Apply the same maintenance-window and SP-blocklist rules used by other SP jobs.
+3. Create an `AbortController` using `DATA_SET_TERMINATION_JOB_TIMEOUT_SECONDS`.
+4. Read `MIN_NUM_DATASETS_FOR_CHECKS` and base dataset metadata.
+5. Scan slots from `minDataSets - 1` down to `DATA_SET_TERMINATION_MIN_INDEX`. For each slot:
```
BigLep: Is the ordering of the slots deterministic? If so, that would mean some datasets will be long lived and some will churn. I haven't thought about this too deeply and appreciate your input, but on the surface, I would imagine doing this with random dataset selection. If we go random, then I assume we do away with `DATA_SET_TERMINATION_MIN_INDEX` approach.

Another way to look at my question is: why are we doing `DATA_SET_TERMINATION_MIN_INDEX` rather than just randomly selecting a dataset?

docs/data-set-termination.md
```diff
@@ -0,0 +1,220 @@
...
+   - a. Build its metadata using the same logic as `data_set_creation`.
+   - b. Classify it via `getDataSetProvisioningStatus()`.
+   - c. Skip if `missing` — nothing to terminate.
+   - d. Skip if `terminated` — `data_set_creation` owns repair of these slots.
+   - e. Skip if `live` but has any deal row with `cleaned_up = false` — the deal job is still tracking it as active.
+6. Call the termination flow on the first slot that passes all skip conditions (reaches step 5e without being skipped).
```
BigLep:
```suggestion
6. Call the termination flow on the first slot that passes all skip conditions (reaches step 5e without being skipped). 
```

docs/data-set-termination.md
```diff
@@ -0,0 +1,220 @@
...
+7. Log the outcome and exit for this tick.
+8. If no eligible slot is found after the full scan, log `skipped.no_candidate` and exit. This is expected when `data_set_creation` has not yet replenished a previously terminated slot.
+
+As with `data_set_creation`, the job performs **at most one state-changing action per invocation**.
```
BigLep:
```suggestion
As with `data_set_creation`, the job performs **at most one state-changing action per invocation**. The way to increase the dataset termination rate is to increase `DATASET_TERMINATIONS_PER_SP_PER_HOUR`.
```

docs/data-set-termination.md
```diff
@@ -0,0 +1,220 @@
...
+
+- If `pdpEndEpoch !== 0` is already set when the job starts (slot is already terminated), skip the `terminateService` call and treat the run as a no-op success.
+- If `terminateService` reverts with an already-terminated error (for example, `"service already terminated"` or `"dataset not active"`), treat it as idempotent success and proceed to the polling step.
+- If the transaction confirms but `pdpEndEpoch` does not become non-zero before the abort signal fires, treat the run as `failure.timedout` and let pg-boss retry on the next tick.
+
+### Metrics and BetterStack dashboards
+
+The termination job has two distinct observability concerns: is the trigger firing, and is the canary signal it produces showing up in creation metrics. Creation metrics are the primary signal; termination metrics are only there to confirm the trigger is working.
+
+All metrics carry the standard label set defined in [`checks/events-and-metrics.md`](./checks/events-and-metrics.md#metrics):
```
BigLep: Do we maybe put all the metrics in events-and-metrics so its all in one place?

docs/data-set-termination.md
BigLep, line 1: High level: should we maybe combine this done with data-set-creation.md ? Maybe there's a data-set-lifecycle.md doc we want?

I haven't thought too deeply about this idea, but I'm basically wondering about whether it's easier to think about the dataset lifecycle as a whole rather than spanning two documents.

docs/data-set-termination.md
```diff
@@ -0,0 +1,220 @@
...
+
+`data_set_termination` only terminates datasets that correspond to dealbot-managed metadata slots. `data_set_creation` detects the resulting `missing` slot through its normal metadata lookup and recreates it without needing any new cross-job state.
+
+Expected healthy behavior:
+
+1. `data_set_termination` calls `terminateService` and polls until `pdpEndEpoch !== 0`.
+2. `data_set_creation` runs next. The Synapse SDK filters the terminated dataset from metadata lookups, so the slot resolves as `missing` immediately. `data_set_creation` provisions a replacement dataset directly in this run.
+3. Existing creation metrics and alerts resume acting as the canary.
+
+**Rate constraint:** `DATASET_CREATIONS_PER_SP_PER_HOUR` should be **greater than or equal to** `DATASET_TERMINATIONS_PER_SP_PER_HOUR`. If termination runs faster than creation, the missing-slot backlog accumulates and the system stops behaving like a simple steady-state canary. The scheduler should emit a startup warning log when this constraint is violated so the misconfiguration is visible without a dashboard.
```
BigLep:
```suggestion
**Rate constraint:** `DATASET_CREATIONS_PER_SP_PER_HOUR` should be **greater than or equal to** `DATASET_TERMINATIONS_PER_SP_PER_HOUR`. If termination runs faster than creation, the missing-slot backlog accumulates and the system stops behaving like a simple steady-state canary. The scheduler emits a startup warning log when this constraint is violated so the misconfiguration is visible without a dashboard.
```

I changed the wording because I'm assume we'll implement it this way.

docs/data-set-termination.md
```diff
@@ -0,0 +1,220 @@
...
+
+**Canary window size:** The number of slots eligible for termination is `MIN_NUM_DATASETS_FOR_CHECKS - DATA_SET_TERMINATION_MIN_INDEX`. A canary window of `1` means a single slot cycles continuously; a larger window gives termination more candidates when one slot has active deals blocking it. In practice, a window of `2`–`3` is usually enough buffer.
+
+## Open Questions
+
+### Should `terminated` be renamed in `getDataSetProvisioningStatus`?
+
+The `terminated` status returned by `getDataSetProvisioningStatus` means: the Synapse SDK resolved a `dataSetId` from the metadata fingerprint but liveness probes failed. This is distinct from a dataset that has `pdpEndEpoch !== 0` on-chain (which the SDK filters out entirely, causing the slot to resolve as `missing`).
+
+The name `terminated` is already used for both the on-chain lifecycle concept and this SDK liveness-probe failure state, which causes confusion. Candidate replacements: `irrecoverable` or `missing.sp`. This rename would affect `data_set_creation`'s handler and repair path as well.
```
BigLep: I don't know what names are used where and the specifics, but lets not make things confusing. A "terminated" dataset already has a definition (and most recently getting updated in https://app.notion.com/p/filecoindev/Data-Set-Terminations-Clean-up-360dc41950c1801ebf0aff017a322e7 I believe). 

If our terminated definition isn't in sync with other understandings of terminated, then lets not use that word.

I don't really know yet what to suggestion in terms of name.

When you say "liveness probes failed", what do you mean? Basically to help with naming, today when `getDataSetProvisioningStatus` returns `terminated` what is happening with the dataset? How could that have happened to the dataset? That is useful info for coming up with a new name here.

But in general: I'm in favor of fixing this and making it more clear and simpler to reason about. Overlaoding the term "terminated" I think moves us in the wrong direction.

docs/data-set-termination.md
```diff
@@ -0,0 +1,220 @@
...
+
+### Should `data_set_termination` absorb the repair path from `data_set_creation`?
```
BigLep: It's good to raise it. I don't know if there is an obvious clear-cut answer.

It sounds like the repair case still involves creating a dataset right, which is why it's reasonable to be in `data_set_creation`.

Maybe we just have an FAQ item for "Why doesn't data-set-termination handle repair?" For the answer we can say it's not clear cut either direction and that logic first lived in create_data_set. There hasn't been a compelling reason to change this.

docs/data-set-termination.md
```diff
@@ -0,0 +1,220 @@
...
+
+### What happens on-chain after `terminateService` is called?
+
+`terminateService` does not delete a dataset instantly. It starts a multi-step on-chain sequence that plays out over roughly 30 days. Understanding this is important because the termination job only needs the first step to complete before it can exit and let `data_set_creation` replenish the slot.
+
+**Step 1 — terminateService tx confirms**
+
+`terminateService` calls `FilecoinPay.terminateRail(pdpRailId)`, which sets `endEpoch = block.number + lockupPeriod` on the PDP rail. The FWSS `railTerminated` callback fires in the same transaction, stores `info.pdpEndEpoch`, and emits `PDPPaymentsTerminated` and `ServiceTerminated`.
+
+This is the point the termination job polls for: `pdpEndEpoch !== 0`. Once this is set, `data_set_creation` will classify the slot as `missing` and begin the replenishment sequence. The termination job's work is done here.
```
BigLep: > termination job polls for: `pdpEndEpoch !== 0`.

Maybe quickly say what happens after this polling when ``pdpEndEpoch !== 0` is found (or you can link to the document where this is described more).

> Once this is set

I think it's not fully obvious what the "this" pronoun is referring to. To make it clear, maybe remove the pronoun and replace with the actual noun?

BigLep (commented):
I'm happy to take another look 2026-06-04, but hpefully this gives enough direction to give confidence about starting implementation.

(This now concludes the review I started with https://github.com/FilOzone/dealbot/pull/588#pullrequestreview-4421331742)

docs/data-set-termination.md
silent-cipher: When `getDataSetProvisioningStatus` return `terminated` it means data set is live/healthy on-chain but curio's `addPieces` path returns `unrecoverable_proving_failure_epoch` state, where SP refuses `addPieces` with HTTP 409.

Ref - https://github.com/FilOzone/dealbot/issues/545

docs/data-set-termination.md
silent-cipher: By repair it means the curio's `addPieces` refuses with `unrecoverable_proving_failure_epoch` but the data set is still live on-chain. So, repair path calls `terminateSerivce` which then actually terminates the data set on-chain. Then, in next tick the data set is replenished.

docs/data-set-termination.md
silent-cipher: I thought we keep some slots stable. We don't exercise `dataSetCreation` -> `terminateService` cycle on them because we need deals to perform retrieval tests on them. If we pick random slot from all available slots for terminations we will barely have deals to do retrievals on them.

silent-cipher: I was thinking of making only the partition deterministic, not the selection itself. The lower-index slots (up to `DATA_SET_TERMINATION_MIN_INDEX`) stay stable to produce deals for retrieval testing, while the slots in `[DATA_SET_TERMINATION_MIN_INDEX, MIN_NUM_DATASETS_FOR_CHECKS)` can be picked randomly for the `create → terminate` cycle.
I'll make it more clear in the docs.

docs/environment-variables.md
```diff
@@ -810,6 +838,27 @@ Use this to stagger multiple dealbot deployments that are not sharing a database
 
 ---
 
+### `DATA_SET_LIFECYCLE_CHECK_JOB_TIMEOUT_SECONDS`
+
+- **Type**: `number`
+- **Required**: No
+- **Default**: `600` (10 minutes)
```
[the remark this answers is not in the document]
silent-cipher: will fix once we align on job timeout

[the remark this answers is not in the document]
silent-cipher: changed default value to `600`

docs/checks/data-set-lifecycle-check.md
```diff
@@ -0,0 +1,128 @@
...
+
+Source code links throughout this document point to the current implementation.
+
+For event and metric definitions used by the dashboard, see [Dealbot Events & Metrics](./events-and-metrics.md).
+
+> **Note**: This check calls `terminateService` to start the on-chain termination sequence. It does **not** call `PDPVerifier.deleteDataSet`, which is SP-initiated. See the [FAQ](#what-happens-on-chain-after-terminateservice-is-called) for details on what happens after termination.
+
+## Overview
+
+A "data set lifecycle check" tests the full `createDataSet → terminateService` lifecycle for a storage provider. Dealbot creates an empty throwaway data set and immediately terminates it in the same run. A successful check confirms both the `createDataSet` and `terminateService` paths work correctly on the SP.
```
BigLep: One thought: most of our tools do createDataSetAndAddPiece right? If it possible that createDataSet could succeed but createDataSetAddPiece could fail then I think it's important to do both (where createDataSetAndAddPiece uses a tiny piece).

docs/checks/data-set-lifecycle-check.md
```diff
@@ -0,0 +1,128 @@
...
+  CreateDataSet -->|abort signal| Fail
+```
+
+### 1. Apply job guards
+
+Dealbot applies the same maintenance-window and SP-blocklist rules used by all other SP jobs. If `DATASET_LIFECYCLE_CHECK_ENABLED` is `false`, the job logs a disabled skip and exits.
+
+### 2. Create the empty data set
+
+Dealbot calls `createDataSet` (from `@filoz/synapse-core/sp`) to create a new empty data set on the SP. The data set is tagged with metadata `{ dealbotLifecycleCheck: "<timestamp>" }`. The fixed `dealbotLifecycleCheck` key is the handle for finding leaked sets later; the per-run value ensures a fresh data set is created on every invocation rather than resolving a prior one.
```
BigLep: What is the "per run" value?
```suggestion
Dealbot calls `createDataSet` (from `@filoz/synapse-core/sp`) to create a new empty data set on the SP. The data set is tagged with metadata `{ dealbotLifecycleCheck: "<timestamp>" }`. The fixed `dealbotLifecycleCheck` key is the handle for finding leaked sets later; the `<timestamp>` value ensures a fresh data set is created on every invocation rather than resolving a prior one.
```

I did this because I was originally confused by what was meant by "per-run value". I see what you're meaning now, but I think this makes it clearer.

docs/checks/data-set-lifecycle-check.md
```diff
@@ -0,0 +1,128 @@
...
+
+This step does **not** emit `dataSetCreation` metrics — those belong to the `data_set_creation` job.
```
BigLep: Turn `data_set_creation` into a hyperlink?

docs/checks/data-set-lifecycle-check.md
BigLep:
```suggestion
A "data set lifecycle check" tests the `createDataSet → terminateService` lifecycle for a storage provider. Dealbot creates an empty throwaway data set and immediately terminates it in the same run. A successful check confirms both the `createDataSet` and `terminateService` paths work correctly on the SP.

Other dataset operations like adding pieces and retrieving them is handled by the [data storage check](LINK). Data set creation isn't covered by the data storage check, and it also relies on a relatively stable set of datasets, which is why this separate check was developed.
```

BigLep (approved):
Main feedback is item is to consider to doing `createDataSet` and `createDataSetAddPiece` (sp - can't remember). I would hate for this canary to pass but then the main operation that users actually use to fail...

docs/checks/data-set-lifecycle-check.md
silent-cipher: I think there can be case where createDataSet could succeed but createDataSetAndAddPiece could fail.

Thought on how do we do it:
1. We choose which path from these two (`createDataSet` | `createDataSetAndAddPieces`) to test via one of the following methods:
    - randomly selecting one of them
    - running both of them in a single job and we reduce job frequency to half
    - selecting alternatively in consecutive job runs
2. We've separate `checkType` in lifecycle metrics - `dataSetLifecycleCheck` and `dataSetWithPiecesLifecycleCheck`

BigLep: Ok, so if we think createDataSet could succeed but createDataSetAndAddPiece could fail, then I think we execute both paths in this lifecycle job.

I think we running both of them in a single job. I don't even know if it matters that we change the default frequency. (I don't really care about the default fequency. I do want to make sure we're running frequently enough in Calibration so each provider has created at least one or two datasets in a 24 hour period.

Do we even need a different checkType. checkType=dataSetLifecycleCheck can just mean that we did both checks in parallel and success means they both completed successfullyl in time and the duration is how long it took to complete both in parallel.

silent-cipher: Agreed, we run both of them in a single job.

My only concern is observability when one succeeds and the other fails. If we want to keep `checkType=dataSetLifecycleCheck`, could we at least add a path label (values - `createDataSet` and `createDataSetAndAddPieces`) to distinguish between them?

BigLep: I'm not too worried about being able to differentiate metrics between `createDataSet` and `createDataSetAndAddPieces`. If we start seeing failures or slowdown with the overall check, I think it's fine to starting to manually digging in to dealbot logs to learn more. 

This is coming from me feeling like our metrics are getting fairly complicated or at least extensive and I only want to add if someting is really needed.

But I don't have a strong opinion here.

docs/checks/data-set-lifecycle-check.md
```diff
@@ -0,0 +1,212 @@
...
+
+Source code links throughout this document point to the current implementation.
+
+For event and metric definitions used by the dashboard, see [Dealbot Events & Metrics](./events-and-metrics.md).
+
+> **Note**: This check calls `terminateService` to start the on-chain termination sequence. It does **not** call `PDPVerifier.deleteDataSet`, which is SP-initiated. See the [FAQ](#what-happens-on-chain-after-terminateservice-is-called) for details on what happens after termination.
+
+## Overview
+
+A "data set lifecycle check" tests the `createDataSet → terminateService` lifecycle for a storage provider. Each tick both creation path variants run in parallel, then each immediately terminates its throwaway data set. Running both variants every tick means both SP code paths are exercised on every invocation.
```
BigLep:
```suggestion
A "data set lifecycle check" tests the `createDataSet|createDataSetAndAddPieces → terminateService` lifecycle for a storage provider. Each tick _both_ creation path variants run in parallel, then each immediately terminates its throwaway data set.
```

docs/checks/data-set-lifecycle-check.md
```diff
@@ -0,0 +1,212 @@
...
+
+**Empty variant**: exercises the `createDataSet` path.
+1. Creates a new empty data set, tagged with `dealbotLifecycleCheck` metadata
+2. Waits for the SP to confirm the data set on-chain (`dataSetId` returned)
+3. Calls `terminateService` on the created data set and waits for the transaction receipt
+
+**With-pieces variant**: exercises the `createDataSetAndAddPieces` path.
+1. Uploads a small deterministic canary piece to the SP's HTTP storage service
```
BigLep: Why deterministic? I actually think random (e..g., use current timestamp) is better to make sure that we aren't just succeeding because of piece caching.

docs/checks/data-set-lifecycle-check.md
```diff
@@ -0,0 +1,212 @@
...
+5. Calls `terminateService` on the created data set and waits for the transaction receipt
+
+A successful check requires all steps of **both** variants to complete within the allowed time. If either variant fails, the overall job fails. If both variants fail, an `AggregateError` is thrown. Failure occurs if any step fails or the check exceeds `DATA_SET_LIFECYCLE_CHECK_JOB_TIMEOUT_SECONDS`.
+
+The two variants emit metrics under distinct `checkType` labels (`dataSetLifecycleCheck` and `dataSetWithPiecesLifecycleCheck`) so dashboards can track them independently.
+
+## What Gets Asserted
+
+Each data set lifecycle check asserts the following for every SP. Both variants run every tick in parallel. Assertions 1–3 apply to the empty variant; assertions 1′–5′ apply to the with-pieces variant. Both paths share the timeout assertion.
+
```
BigLep:
```suggestion
```

I think you can just remove to keep this more concicse

docs/checks/data-set-lifecycle-check.md
```diff
@@ -0,0 +1,212 @@
...
+| 4′ | Data set and piece are confirmed on-chain | `waitForCreateDataSetAddPieces` resolves with a `dataSetId` and `piecesIds` | [`dataSetWithPiecesLifecycleCheckStatus`](./events-and-metrics.md#dataSetWithPiecesLifecycleCheckStatus) |
+| 5′ | `terminateService` succeeds on the created data set | `terminateServiceSync` call completes and the transaction receipt is received | [`dataSetWithPiecesLifecycleCheckMs`](./events-and-metrics.md#dataSetWithPiecesLifecycleCheckMs) |
+| 6′ | All steps complete within the timeout | Check is not marked successful until all steps pass within `DATA_SET_LIFECYCLE_CHECK_JOB_TIMEOUT_SECONDS` | [`dataSetWithPiecesLifecycleCheckMs`](./events-and-metrics.md#dataSetWithPiecesLifecycleCheckMs) |
+
+## Data Set Lifecycle Check Lifecycle
+
+The dealbot scheduler triggers data set lifecycle check jobs at a configurable rate. On each tick, both creation variants run in parallel via `Promise.allSettled`.
+
+```mermaid
+flowchart TD
```
BigLep, line 58: This works and is accurate, and I don't think it's necessarily worth changing, but I think you could simplify it by removing/deduplicating these boxes:
- "Run both variants in parallel"
- terminateServiceSync
- Record success
- Record failure
- Collect results

I know that isn't accurate but it would reduce on the number of boxes and lines. It kind of matches how down below you have "5. Terminate the service (both variants)"

Anyways, something to consider for future

docs/checks/data-set-lifecycle-check.md
```diff
@@ -0,0 +1,212 @@
...
+
+### 3b. Wait for data set creation confirmation
+
+Dealbot calls `waitForCreateDataSet` with the `statusUrl` returned by the SP. When the SP confirms the data set is created on-chain, it resolves with a `dataSetId`.
+
+### 4. With-pieces variant: exercise `createDataSetAndAddPieces` path
+
+### 4a. Upload canary piece
+
+Dealbot calls `uploadPieceStreaming` to push a small fixed canary piece (256 bytes, all `0x61`) to the SP's HTTP storage service. The piece is deterministic so a leaked data set can always be identified by its piece CID alongside the `dealbotLifecycleCheck` metadata key.
```
BigLep: Per above, I'm not sure there is value using the same pieceCID for all of these.

docs/checks/data-set-lifecycle-check.md
```diff
@@ -0,0 +1,212 @@
...
+
+Dealbot calls `waitForCreateDataSetAddPieces` with the `statusUrl` returned by the SP. When the SP confirms both the data set creation and piece addition on-chain, it resolves with `{ dataSetId, piecesIds }`.
+
+### 5. Terminate the service (both variants)
+
+Dealbot calls `terminateServiceSync` (from `@filoz/synapse-core/warm-storage`) on the newly created `dataSetId`. This submits the terminate transaction and waits for the receipt, confirming the termination was recorded on-chain. This is Step 1 of the [full on-chain termination sequence](#what-happens-on-chain-after-terminateservice-is-called). The job does not wait for the full ~30-day rail finalization.
+
+The entire check (all variant steps + termination) is bounded by `DATA_SET_LIFECYCLE_CHECK_JOB_TIMEOUT_SECONDS`. A timeout is classified as `failure.timedout`.
+
+## Check Status Progression
```
BigLep, line 118: I'm not seeing a ton of value of separating out the status progression across the two independent paths. I think you can make this one table and not worry about differentiating just to keep it simple. This is coming from the regard of wanting to keep it simple (less text to read, less code to maintain, less events/metrics to dashboard)

docs/checks/data-set-lifecycle-check.md
```diff
@@ -0,0 +1,212 @@
...
+
+**With-pieces variant** — recorded via [`dataSetWithPiecesLifecycleCheckStatus`](./events-and-metrics.md#dataSetWithPiecesLifecycleCheckStatus):
+
+| Overall Status | Meaning |
+|--------|---------|
+| `success` | All steps passed: canary piece uploaded and ingested, data set created with piece, confirmed on-chain, service terminated. |
+| `failure.timedout` | The job was aborted because it exceeded `DATA_SET_LIFECYCLE_CHECK_JOB_TIMEOUT_SECONDS`. |
+| `failure.other` | Any other failure: upload failed, `findPiece` timed out, `createDataSetAndAddPieces` failed, or `terminateService` failed. |
+
+## Metrics Recorded
```
BigLep, line 128: Similar, I don't think we need metrics for the individual variants.

docs/checks/data-set-lifecycle-check.md
```diff
@@ -0,0 +1,212 @@
...
+
+**Step 3 — data set deletion at PDPVerifier (SP-initiated).** After the rail finalizes, the SP may call `PDPVerifier.deleteDataSet`. The lifecycle check does not wait for steps 2 or 3 — waiting ~30 days per invocation would defeat the purpose of a canary.
+
+### Why two variants?
+
+The data storage check (`data_set_creation` job) uses `createDataSetAndAddPieces` internally when creating a new data set with a piece. A canary that only exercises `createDataSet` would leave `createDataSetAndAddPieces` untested by the lifecycle check.
+
+Both variants run on every tick. It ensures both SP code paths are exercised on every invocation rather than in expectation over two ticks. The `Promise.allSettled` approach also ensures that a failure in one variant is never hidden by success in the other.
+
+The empty variant exercises the `createDataSet → waitForCreateDataSet → terminateService` path. The with-pieces variant exercises `uploadPieceStreaming → findPiece → createDataSetAndAddPieces → waitForCreateDataSetAddPieces → terminateService`. The two `checkType` label values (`dataSetLifecycleCheck` and `dataSetWithPiecesLifecycleCheck`) let Grafana track them as independent time series.
```
BigLep: Potentially revise if we simplify and don't have checkType.

docs/checks/data-set-lifecycle-check.md
```diff
@@ -0,0 +1,212 @@
...
+
+### What if creation succeeds but termination fails?
+
+If creation succeeds but termination fails (process crash, job timeout, or an on-chain error that is not an already-terminated no-op), the created data set stays live on the SP. This is called a leak and is an accepted trade-off for keeping the job self-contained.
+
+Leaked sets are discoverable by filtering data sets with the `dealbotLifecycleCheck` metadata key — this key is set by both variants. For the with-pieces variant, the canary piece is additionally identifiable by its fixed piece CID (256 bytes, all `0x61`). Each leak is recorded in the log line (message: "throwaway data set may have leaked") with the `dataSetId` included for easy identification.
```
BigLep: Potentially revise if we start having a random tiny piece.

docs/checks/events-and-metrics.md
```diff
@@ -100,7 +100,7 @@ sequenceDiagram
 * They are exported via Prometheus.
 * All Prometheus/OpenTelemetry metrics have label/attributes for:
    - `network=calibration|mainnet`
-   - `checkType=dataStorage|retrieval|dataRetention|dataSetCreation|pullCheck` — attribute metrics to a particular check/job
+   - `checkType=dataStorage|retrieval|dataRetention|dataSetCreation|dataSetLifecycleCheck|dataSetWithPiecesLifecycleCheck|pullCheck` — attribute metrics to a particular check/job
```
BigLep: Adjust here and elow if we just have a singular dataSetLifecycleCheck

docs/environment-variables.md
```diff
@@ -676,6 +676,34 @@ rate-based (per hour) and persisted in Postgres so restarts do not reset timing.
 
 ---
 
+### `DATASET_LIFECYCLE_CHECK_ENABLED`
+
+- **Type**: `boolean`
+- **Required**: No
+- **Default**: `true` on calibration, `false` on mainnet
```
BigLep:
```suggestion
- **Default**: `false` on mainnet, `true` everywhere else
```

this way we account for devnets

that is how I'd implement it as well

docs/environment-variables.md
```diff
@@ -810,6 +838,27 @@ Use this to stagger multiple dealbot deployments that are not sharing a database
...
+- **Enforced**: Yes (config validation, effective floor applied at runtime)
+
+**Role**: Maximum runtime for `data_set_lifecycle_check` jobs before forced abort via `AbortController`. Both variants run in parallel within this budget — the timeout bounds the wall-clock time from job start until both variants have settled (or been aborted).
+
+**When to update**:
+
+- Increase if the with-pieces variant (`uploadPieceStreaming` + `findPiece` + `createDataSetAndAddPieces` + `terminateServiceSync`) consistently times out on slow networks, since it has more steps than the empty variant and will typically be the critical path.
+- Decrease for faster fail-fast behavior during testing.
+
+**Note**: If the configured value is below 60 seconds, the runtime silently raises it to 60 seconds as an effective floor. An abort due to this timeout (or an internal poll timeout) is recorded as `dataSetLifecycleCheckStatus{value="failure.timedout"}` (empty variant) or `dataSetWithPiecesLifecycleCheckStatus{value="failure.timedout"}` (with-pieces variant) and retried on the next scheduled tick.
```
BigLep: Simplify if we don't differentiate between the variants.

BigLep (approved):
Thanks for iterating!

SgtPooki (approved):
looks good to me, lets get this out on calibration.. going to wait for images to build from #571 before I merge this and interrupt the image builds though

PR #602, opened 2026-06-12
refactor: migrate dataset termination to provider-relayed flow

### LINKED ISSUE #546: Dealbot can't auto-terminate datasets when payer is a Safe multisig
## Problem

Dealbot cannot auto-terminate PDP datasets when the payer is a Safe multisig. The repair path added in #518 (`data_set_creation` → `repairTerminatedDataSet` → `synapse.storage.terminateDataSet`) reverts at the simulate step with `CallerNotPayerOrPayee(caller=0x0)`.

## Why

1. We run in session-key mode with a Safe multisig owner. `apps/backend/src/common/synapse-factory.ts` installs a `safeReadTransport` (PR #419) that strips `from` from every `eth_call`/`eth_estimateGas`. Reason: Lotus rejects `eth_call` with `from = contract` ([lotus#13470](https://github.com/filecoin-project/lotus/pull/13470) still open).
2. `synapse-sdk` `WarmStorageService.terminateDataSet` calls `terminateService(this._client, ...)` and synapse-core does `simulateContract` then `writeContract` on the same client. Simulation runs through the strip-from read client, contract sees `msg.sender = 0x0`, reverts.
3. Bypassing simulation does not help today: `terminateService` is gated by `msg.sender ∈ {payer, payee}` and was not in `DefaultFwssPermissions`, so a session-key signed write reverts at execution.

FilOzone/synapse-sdk#796 renames the permission to `TerminateServicePermission` and adds `signTerminateService`. It does not yet wire the signed extraData into the contract call, and it does not address the strip-from interaction.

## Impact

Terminated datasets at any SP we use have to be cleaned up by hand via Safe Transaction Builder, as in #545 (197 datasets, 9 multisig batches). USDFC lockup keeps accruing until we do.

## Next step

Track until the auto path works end to end. We need:

- FWSS contract entry point that accepts a session-key-signed `extraData` for termination
- synapse-sdk PR that routes `terminateDataSet` through `sessionClient` and signs with `signTerminateService`
- Dealbot: re-mint the session key once the new permission is required
- Either the new entry point auths purely via sig recovery (so `msg.sender = 0x0` is acceptable during simulate), or relax `safeReadTransport` for that method, or wait for filecoin-project/lotus#13470 to land

## Refs

- #545 manual cleanup for provider 24
- #518 `data_set_creation` repair handler
- #537 multi-probe liveness detection
- #419 `safeReadTransport` workaround
- FilOzone/synapse-sdk#796
- filecoin-project/lotus#13470

silent-cipher:
Closes #546 

**Note:** Need `filecoin-pin` utilizing `synapse-sdk^1.0.1` and `synapse-core^0.7.0`.

### DIFF
--- a/apps/backend/src/data-set-lifecycle/data-set-lifecycle.service.ts
+++ b/apps/backend/src/data-set-lifecycle/data-set-lifecycle.service.ts
@@ -2,11 +2,13 @@ import {
   createDataSet,
   createDataSetAndAddPieces,
   findPiece,
+  type TerminateServiceStatusSuccess,
+  terminateService,
   uploadPieceStreaming,
   waitForCreateDataSet,
   waitForCreateDataSetAddPieces,
+  waitForTerminateService,
 } from "@filoz/synapse-core/sp";
-import { terminateServiceSync } from "@filoz/synapse-core/warm-storage";
 import { Injectable, Logger } from "@nestjs/common";
 import { awaitWithAbort } from "../common/abort-utils.js";
 import { type ProviderJobContext, toStructuredError } from "../common/logging.js";
@@ -22,6 +24,24 @@ type LifecycleBaseLogContext = {
   providerId: bigint;
   providerName: string;
 };
+
+/**
+ * Provider-relayed termination as a single call: sign + POST the EIP-712
+ * authorization (`terminateService`) then poll until the SP confirms it on-chain
+ * (`waitForTerminateService`). Replaces the old warm-storage `terminateServiceSync`,
+ * which submitted the tx directly — a path that reverts for Safe-multisig payers
+ * because the session-key signer is never the on-chain msg.sender (#546).
+ */
+async function terminateServiceSync(
+  client: SynapseViemClient,
+  options: { dataSetId: bigint; serviceURL: string; onHash?: (hash: `0x${string}`) => void },
+): Promise<TerminateServiceStatusSuccess> {
+  const { statusUrl } = await terminateService(client, {
+    dataSetId: options.dataSetId,
+    serviceURL: options.serviceURL,
+  });
+  return waitForTerminateService({ statusUrl, onHash: options.onHash });
+}
 
 @Injectable()
 export class DataSetLifecycleService {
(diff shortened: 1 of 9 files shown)


apps/backend/src/data-set-lifecycle/data-set-lifecycle.service.ts
```diff
@@ -23,6 +25,24 @@ type LifecycleBaseLogContext = {
...
+async function terminateServiceSync(
+  client: SynapseViemClient,
+  options: { dataSetId: bigint; serviceURL: string; onHash?: (hash: `0x${string}`) => void },
+): Promise<TerminateServiceStatusSuccess> {
+  const { statusUrl } = await terminateService(client, {
+    dataSetId: options.dataSetId,
+    serviceURL: options.serviceURL,
+  });
+  return waitForTerminateService({ statusUrl, onHash: options.onHash });
+}
```
SgtPooki, line 44: nit: `terminateServiceSync` implies a non-async function here, but we can see that it's async. 

non-nit:

`@filoz/synapse-core/sp.terminateService()` throws on provider 409 responses:
- `ServiceAlreadyTerminatedError` when the SP reports the data set is already terminated
- `TerminateServicePendingError` when a termination request is already queued

The SDK helper catches those and either returns the known `endEpoch` or resumes polling via `terminateServiceStatusUrl(...)`. This local lifecycle helper calls the lower-level API directly, so a retry/duplicate lifecycle run can fail the canary even though the throwaway data set is already terminated or termination is still pending.

Can use the SDK termination flow here?

silent-cipher: My initial thought was that, since this helper exercises dataset creation and termination within a single job run, the chances of hitting `ServiceAlreadyTerminatedError` or `TerminateServicePendingError` seem quite low.

The SDK termination flow would also introduce ~6-7 additional RPC calls (dataset resolution, debt checks, etc.) that may not be necessary here.

That said, if we're not concerned about the extra RPC overhead, I'm happy to switch this to the SDK termination flow via `StorageManager`. Let me know what you think.

SgtPooki: ok i like your thinking.. we definitely need to be thinking about RPC optimizations here.

BigLep: @silent-cipher : I'm not up on the specifics here, but this seems like important info to have a comment for so our furture selves or a future agent doesn't come through and try and "simplify".

